PR #22, opened 2014-12-29
Initial work on moving the query to the server

paf31:
/cc @hdgarrood 

This isn't ready yet. I still need to work on the client-side changes, which will probably be considerable, but this implements the bare-bones server-side functionality needed for the new architecture.


paf31:
Ok, I've set up the community server to pull the library list from GitHub every night at midnight. It's now running here:

http://pursuit.purescript.org/

I'm going to fix the couple of things that you mentioned, as well as integrating the query string on page load, and then I'm going to merge this, and forward purescript.org/pursuit to the new URL.

michaelficarra:
The first thing I noticed is that the search is not auto-focused as it used to be.

paf31:
@michaelficarra Yep, that's on my list too :)

PR #30, opened 2015-01-29
Added a bunch of libraries.

joneshf:
Finally got around to adding some of my libraries. Let me know if I did this incorrectly.

Also, now that things have changed, what's the prefered way of adding a library? How can we test that things didn't break? I run `cat libraries.json | ./dist/build/pursuit-gen/pursuit-gen -o data.json` and then `./dist/build/pursuit-server/pursuit-server` and can then curl stuff, so I assume it works. But, I can't figure out how the client is supposed to work with that.

### DIFF
--- a/server/libraries.json
+++ b/server/libraries.json
@@ -1,15 +1,43 @@
 [
   {
+    "bowerName": "purescript-ajax",
+    "gitUrl": "https://github.com/joneshf/purescript-ajax"
+  },
+  {
+    "bowerName": "purescript-algebra",
+    "gitUrl": "https://github.com/joneshf/purescript-algebra"
+  },
+  {
+    "bowerName": "purescript-angle",
+    "gitUrl": "https://github.com/joneshf/purescript-angle"
+  },
+  {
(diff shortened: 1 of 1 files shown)


paf31:
Let's see... :)

paf31:
I just ran the job by hand, and it looks like it's working.

http://pursuit.purescript.org/?(.~)

/cc @hdgarrood

hdgarrood:
Maybe we could have a Travis CI job that attempts the first of the above commands? If the it exits with 0, I think there's a pretty good chance it's been done correctly.

paf31:
Well, the script already bails out if the generator fails, so I think it's doing the same job, but CI couldn't hurt.

hdgarrood:
We can ask Travis to run builds on each PR so this could work rather nicely

hdgarrood:
Yes - the benefit with Travis would be that a) the submitter gets a notification that the build failed, and b) there's a much lower chance of jobs on the VM failing because of a mistake.

joneshf:
Thanks!

Issue #6, opened 2014-08-11
Include repo name and url in results


natefaubion:
So its more obvious which entry to add to a bower file.

hdgarrood:
I was thinking perhaps changing the structure of libraries.json slightly might make this easier - if we have entries in libraries.json look like this:

```
{
  "bowerName": "purescript-arrays",
  "github": { "user": "purescript", "repo": "purescript-arrays" }
}
```

then we can more easily generate a HTTPS url that's suitable for use in the pursuit UI. The current format, which uses a key called `gitUrl`, implies that a url with a different scheme would also work, like "git:"

If we do want to include a library that isn't on github, then we can extend this later so that an entry may have either a `github` key, or a `gitUrl` key (which is just passed straight to `git clone`, as happens currently).

paf31:
I like that idea.

Also, I was thinking a lot about moving the data into a Postgres instance on the backend, maybe using something like groundhog to do the mapping. This way, we can avoid having an ever-growing JSON file, and also add new things to the schema, like examples, markdown docs, a page for each library, etc.

hdgarrood:
Sounds good - so would libraries.json be replaced with a Haskell source file, that gets loaded into the database periodically?

paf31:
I think we can keep `libraries.json`, it's just that during the `git clone` job, we put the data into Postgres instead of generating `data.json`.

Maybe it also makes sense to have each repo include an optional `.pursuit.json` file, with Pursuit-related metadata, like a package description in Markdown format, or links to other Pursuit packages. What do you think?

hdgarrood:
Ahh, ok. I haven't used haskell with a relational database before, should be good :)

Would we use the data in `.pursuit.json` in the Pursuit UI?

paf31:
Yes, that's the idea. It would be like a `.travis.yml` file, so the user wouldn't need to send a PR to update their library's page. But adding or removing a library would still require a PR.

hdgarrood:
Is there a benefit to doing it that way over just linking to a github repository from Pursuit? I would expect people to put that kind of information in their project's README.

PR #35, opened 2015-02-17
Include library web urls in generated data

hdgarrood:
This PR updates the structure of libraries.json (as in #6) as well as updating the generator so that each library's web url is available to the client, in the generated `data.json`.

### DIFF
--- a/server/generator/Main.hs
+++ b/server/generator/Main.hs
@@ -89,7 +89,7 @@ generateAllData = do
         let vers = dropWhile (== 'v') vers'
         p $ "selected " ++ name ++ ": " ++ vers
         gitCheckoutTag vers' dir
-        let info = LibraryInfo vers
+        let info = LibraryInfo vers (libraryWebUrl lib)
         buildLibraryDb (libraryBowerName lib) (Just info) dir
 
   preludeDb <- buildPreludeDb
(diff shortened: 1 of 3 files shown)


hdgarrood:
At the moment, the client doesn't use any of this newly available data - I'll look into that after this is merged.

paf31:
This looks good, and I'm happy to merge now.

However, I'm thinking of turning Pursuit into a fully server-side application, combining the current search with the `psc-pages` generated documentation. I feel like this will make it easier to implement things like type-based search.

I was going to create an issue to discuss this today, but got sidetracked, but I would definitely like to discuss this with you.

Given that, I think this is still useful (we'll still want this data available), but I'm not sure how it fits in with a rewrite.

Let me know what you think.

hdgarrood:
Sounds good to me. I'm having a little difficulty picturing how it would all go together, though. Are you picturing something like Hackage? Would the process of adding a library remain the same - that is, a PR modifying libraries.json?

paf31:
Yes, I was thinking that the app would include your Git logic, and the current Cron job would just be a timer which reran the job and updated a `TVar` containing the database or something similar. We could also use Postgres or something similar, but I'm not sure that's necessary.

The app would not create `data.json`, but would still load and process `libraries.json`.

hdgarrood:
Ok, cool. In that case, if I've understood correctly, I think most of the code in this PR wouldn't need to change.

How do you think we should progress now?

paf31:
Let's start an issue to discuss the new design. I'll write something up tomorrow.

I expect this will make things simpler, since we'll only have one code base to work on, but we'll see.

If you're on IRC this week, maybe we can discuss there too.

pelotom:
Just a voice from the peanut gallery, but it would be really cool to have this feature in the deployed version while you're busy refactoring things :grin:

paf31:
I'll deploy this now.

paf31:
@pelotom @hdgarrood This is now running on pursuit.purescript.org.

pelotom:
Thanks! I take it there are still changes needed on the front-end to get links to appear in search results?

paf31:
Yes, but it's pretty simple. I can take a look.

paf31:
Done.

pelotom:
I see the library name now but it doesn't appear to be a link...?

pelotom:
Not that that's really needed, I suppose... this is great! :+1:

paf31:
Well, the client doesn't get the link, right now. I could link to Bower, I guess?

pelotom:
Nah, I don't think that's necessary :)

PR #42, opened 2015-02-28
Add purescript-free-canvas

hdgarrood:
Not only is it a useful library, it also includes an exported type with
non-exported constructors - I want to make sure Pursuit handles this properly.


hdgarrood:
Looks like it does. Hooray!

PR #55, opened 2015-02-28
Switch from String to Text, fixes #45

### LINKED ISSUE #45: Use Text instead of String?
Particularly in `Pursuit.Data` and `Pursuit.Database`. Maybe this would fix the issue we were having with running pursuit-server under upstart? Either way it seems like a good idea.

Also I guess since most of these strings are pretty small (module names, package names, type signatures) then we should be using strict Text?

hdgarrood:
DeclDetail is still a lazy Text. But I guess that's not worth worrying about too much for now, especially if we're going to implement type system search and/or switch to Postgres?

### DIFF
--- a/server/PursuitServer/Server.hs
+++ b/server/PursuitServer/Server.hs
@@ -6,7 +6,6 @@ module PursuitServer.Server where
 
 import Data.Monoid
 import qualified Data.Text.Lazy as TL
-import qualified Data.Text as T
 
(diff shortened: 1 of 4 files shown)


hdgarrood:
Oh - wait - apparently I should be using `T.toLower`, not `T.map toLower`

paf31:
:+1: Thanks!

PR #57, opened 2015-02-28
Search with and without parentheses. Fixes #28

### LINKED ISSUE #28: searching for symbolic names no longer returns any results
I tried searching for `>>=`, but no results were returned. This used to work.

hdgarrood:
`@+` from Data.IxSet returns results that match any of the values in the argument list.


paf31:
:+1:

PR #62, opened 2015-03-12
better logging + another test

hdgarrood:


hdgarrood:
@paf31 the live version of pursuit doesn't appear to be including purescript-lists still, perhaps updating it with this code will give a little insight as to what might be happening?

Issue #49, opened 2015-02-28
Use GitHub API instead of directly using Git


paf31:
All we really need to do is

- Directory listings
- Reading files

which are covered by, for example:

- https://api.github.com/repos/purescript/purescript-arrays/contents
- https://raw.githubusercontent.com/purescript/purescript-arrays/master/bower.json

This way we can avoiding using Git on the command line, and only pull down the files we really need (`bower.json` and `src`)

paf31:
In fact, [this](http://hackage.haskell.org/package/github-0.13.1/docs/src/Github-Repos.html#contentsFor) seems to do everything we need.

hdgarrood:
We also need tags, but that package does the job with `tagsFor`.

I just tried this but `contentsFor` is failing with a JsonError because of a missing "encoding" key. I wonder if the github api has changed since this library was written...

hdgarrood:
Ah, no, looking at the GitHub API https://developer.github.com/v3/repos/contents/ - the authors of that package haven't handled the case where the result is a directory properly.

hdgarrood:
I've sent a PR that handles this case, so this issue is now blocked on jwiegley/github#95

hdgarrood:
No longer an issue, since package authors submit their own packages now.

PR #64, opened 2015-03-16
Various improvements

### LINKED ISSUE #1: results include values that are not exported
For example, a search for "regex" includes both `regex` and `regex'`, even though `regex'` [is not exported](https://github.com/purescript/purescript-strings/blob/4bd42a1c92d09ae789d485fbce1d1ffc0e7f7037/src/Data/String/Regex.purs) in `Data.String.Regex`.

hdgarrood:
* Partially switch to GitHub API rather than shelling out to git. This is just for retrieving tags, so far (refs #49)
* Include data constructors in the generated database
* Ensure only exported values are included (fixes #1)
* Fix some warnings in `Pursuit.Docs`

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -31,7 +31,8 @@ library
                    ixset -any,
                    blaze-html -any,
                    cheapskate -any,
-                   data-default -any
+                   data-default -any,
+                   github -any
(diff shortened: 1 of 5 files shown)


paf31:
Thanks!

Issue #48, opened 2015-02-28
Parallelize Generator


hdgarrood:
I'm pretty sure the current generator works sequentially, and spends a lot of time blocked on IO. If we build each package's database in parallel we might be able speed up the generator significantly.

hdgarrood:
No longer relevant; the generator doesn't exist any more.

PR #65, opened 2015-03-17
Update for purescript-profunctor + concurrent database generation

hdgarrood:
It seems `purescript-profunctors` has been renamed to `purescript-profunctor`.

I've also addressed #48 and #49 too, to some extent, by making the generator work concurrently. This necessitated adding a `ReaderT GithubAuth` to the `Generate` monad, so that it's possible to use the GitHub API with authentication (because of rate limiting). You pass an OAuth token in to the server via a `GITHUB_AUTH_TOKEN` environment variable.

Now, it only takes about 10 seconds to build the whole database!

However, I don't think this is quite ready to merge. Problems:

* I got a `TimeoutTriggered` exception once, I think this was because GitHub didn't respond quickly enough. This might be because it opens as many concurrent connections as there are entries in libraries.json. (This might not be the best idea...?)
* For some reason, when I run the server normally: `GITHUB_AUTH_TOKEN="abc... " cabal run` none of the css is served, I just get 404s instead. This doesn't occur if I pass `-l libraries-minimal.json` which is the smaller one, with just one library. Literally no idea whatsoever with this one.
* The code for `mapConcurrently` could potentially be a lot better. It was surprisingly difficult to write that function!

### DIFF
--- a/libraries.json
+++ b/libraries.json
@@ -260,6 +260,10 @@
     "github": { "user": "purescript", "repo": "purescript-validation" }
   },
   {
+    "bowerName": "purescript-vector-space",
+    "github": { "user": "mkaemmerer", "repo": "purescript-vector-space" }
+  },
+  {
     "bowerName": "purescript-virtual-dom",
     "github": { "user": "purescript-contrib", "repo": "purescript-virtual-dom" }
   },


paf31:
Looks nice. Maybe something like `MonadBaseControl` would allow you to simplify `mapConcurrently`?

PR #67, opened 2015-03-19
Update to purescript 0.6.9.3, fix #34

### LINKED ISSUE #34: Generated data has the wrong version for Prelude
In `generator/Main.hs`, we have:
```hs
import qualified Paths_pursuit as Paths
```
and then, further down, in the definition of `buildPreludeDb`:
```hs
  let preludeInfo = LibraryInfo (showVersion Paths.version) preludeWebUrl
```

This is a mistake I made - it's using the version of Pursuit rather than the version of PureScript in the Prelude's `LibraryInfo`. Any ideas how we can get the PureScript version instead? One way could be to add a function to PureScript itself that returns `Paths.version`.

hdgarrood:
Fixes #34 

* Use the correct version of PureScript for Prelude documentation
* Update to PureScript 0.6.9.3, and use the safer `exportedDctors` and
  `exportedDecls` where possible to simplify docs generation.

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -13,7 +13,7 @@ data-dir: ""
 
 library
     build-depends: base ==4.7.*,
-                   purescript >=0.6.7,
+                   purescript >=0.6.9.3,
                    directory >=1.2.0.1 && <1.3,
                    filepath >=1.3.0.1 && <1.4,
                    split -any,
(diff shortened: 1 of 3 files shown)


hdgarrood:
Actually I'll just put all the stuff I've been doing into a single PR.

PR #68, opened 2015-03-19
Various

### LINKED ISSUE #34: Generated data has the wrong version for Prelude

### LINKED ISSUE #43: Documentation for primitive types
Eg:
* http://pursuit.purescript.org/?q=Number
* http://pursuit.purescript.org/?q=String
* http://pursuit.purescript.org/?q=%5B%5D

It would be nice, particularly for beginners, for something to appear for these searches.
Is it possible to include documentation about these in the Prelude? Maybe by re-exporting them?

hdgarrood:
fixes #34, #43 

* Use the correct version of PureScript for Prelude documentation
* Update to PureScript 0.6.9.3, and use the safer `exportedDctors` and
  `exportedDecls` where possible to simplify docs generation.
* Clean up the cabal file a bit
* Add a `Prim` PureScript module, just for documentation purposes
* Cherry-picked the `purescript-profunctor{s =>}` commit because that is ready to go in now (unlike the rest of #65)


prim/Prim.purs
```diff
@@ -0,0 +1,54 @@
...
+
+-- | A JavaScript String.
+-- |
+-- | Construct values of this type with literals:
+-- | * `x = "hello, world" :: String`
+-- |
+-- | Multi-line string literals are also supported with triple quotes (`"""`).
+foreign import data String :: *
+
+-- | A JavaScript Boolean value. The Prelude exports the two values of this
```
paf31: Not technically true, they're built in literals in the compiler, but... eh.

paf31:
Looks great!

prim/Prim.purs
hdgarrood: If I only got one thing wrong then I'm pretty happy with that! I'll fix this in a sec.

PR #73, opened 2015-03-31
Force UTF8 for reading Text files

hdgarrood:
This should hopefully fix the issue where pursuit runs fine under nohup
but breaks under upstart.

This commit replaces the use of Data.Text.IO.readFile with
Data.ByteString.readFile, followed by decodeUtf8, as the Data.Text.IO
documentation recommends in situations where the encoding is known.


paf31:
Thanks!

PR #74, opened 2015-03-31
Optionally authenticate to GitHub API

hdgarrood:
This the same as what's in #65, except that only the code which concerns GitHub API authentication is included - the concurrent stuff is left out.


paf31:
This is running on the server again now - thanks so much for all the work on this. The libraries are also updating now, and the UTF8 issue is fixed. :+1:

hdgarrood:
Hooray!

PR #75, opened 2015-03-31
Fix prim docs

hdgarrood:
Make the docs say that `true` and `false` are literals, not exported by Prelude.

Please merge this after #74


paf31:
:+1: Thanks!

PR #77, opened 2015-03-31
Refactoring + new queries in Pursuit.Database

hdgarrood:
Just a tiny little change that is mergeable now. I'm planning to use the new functions later, for the Hoogle stuff.


paf31:
:+1:

PR #80, opened 2015-04-24
Just stumbled upon datetime not being in libraries.json

eskimor:
Also added arrows. Is it ok, to send such small pull requests on the go or should I collect and make a few bigger pull requests?

Also are there any libraries which should not be included for some reason?


eskimor:
btw. I started building the database locally again, so at least JSON syntax should be fine.

paf31:
:+1: Thanks!

Small or large PRs are both fine.

PR #95, opened 2015-07-04
Time to merge pursuit-2 into master?

hdgarrood:


paf31:
Sounds good, thanks! I'll review this over the weekend.

Once this version is deployed, I can switch the subdomains over if you think it's ready?

michaelficarra:
Is this deployed anywhere currently so we can test it out first?

paf31:
I'm deploying the latest now, but [new-pursuit.purescript.org](http://new-pursuit.purescript.org) has a fairly recent version.

hdgarrood:
I don't think we should switch the subdomains over until at least after 0.7.1 is released, because of https://github.com/purescript/purescript/pull/1221. Also I would like to make psc-publish include its version, so that we can check it on the server and issue warnings if necessary ("please use a newer version").

hdgarrood:
Also new-pursuit probably wants all of its data being regenerated because of that same issue

hdgarrood:
Oh and psc-publish gives a really ugly error if you used `bower link`, I think that needs fixing too

hdgarrood:
I think this code is good to go. See #101 for what I think needs to happen before we switch the subdomains.

paf31:
Could you please add a [suitable license](http://www.google-melange.com/gsoc/document/show/gsoc_program/google/gsoc2014/help_page#2._What_licenses_do_I_have_to_choose) to the `.cabal` file?

hdgarrood:
Ah, of course. Does MIT sound good? That's what all the other purescript code is released under, right?

paf31:
MIT sounds good. Is it compatible with all of the dependencies?

Also:

- A Travis badge in the README file?
- Help link in README goes to a 404 page
- The architecture notes are a bit out of date, and still talk about verification links. Not sure if you want to keep that or merge it into README?

These things can be fixed after the merge obviously.

hdgarrood:
I'll sort these things out, let's merge afterwards.

paf31:
Is `DocsAsHtml.filePathFor` needed still?

paf31:
What's `$logDebug`? I see it in a few places.

hdgarrood:
The dependencies are all either MIT, BSD3, or PublicDomain, apart from cpphs and polyparse, which are GNU LGPL. Those come from hoogle -> haskell-src-exts -> cpphs -> polyparse. I am not 100% sure what the implications of this are; we might have to use LGPL instead. Hoogle is BSD3, though, which would suggest that MIT is ok.

hdgarrood:
`filePathFor` is still used in the documentation index, which is mostly unchanged from how it was in psc-pages, as I haven't got around to implementing an index yet.

`$logDebug` is from [monad-logger](http://hackage.haskell.org/package/monad-logger-0.3.13.2/docs/Control-Monad-Logger.html), and will currently generate a message to standard output (I think); presumably we could configure this behaviour.

That help link in the README will work after the next deploy ;)

hdgarrood:
Having done a little more research, I am pretty sure MIT is fine.

hdgarrood:
The travis build is failing because it doesn't have enough memory to build `haskell-src-exts` :(

hdgarrood:
I think the new Travis infrastructure could probably cope, though: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

paf31:
I think it's probably fine, but I'll do some more reading, and to be safe, we probably want to avoid distributing binaries, attribute clearly, and provide a notice explaining how to modify the LGPL components.

paf31:
One last thing - could you please explain how the session data works? I assume it's stored in a cookie, but I don't know Yesod. Also, I see it contains the OAuth token - does it contain anything else?

I'm happy to merge this whenever you think it's ready.

paf31:
Also, very nice work, FWIW :) This is some of the most readable Haskell code I've reviewed, and as I say, I don't generally read Yesod code.

hdgarrood:
Wow, thank you :)

Shall I copy the script from the compiler repo to generate a big LICENSE? I think that ought to be sufficient re clear attribution. Also, I guess uploading binaries to GH releases via Travis probably counts as distributing binaries? :(

Sessions are using the Yesod default, which puts all the session data into a signed, encrypted cookie, using the `clientsession` package. Currently, the only things that go in there are the logged in user's github username and OAuth token.

Also the new Travis infrastructure can cope with all our dependencies, but currently it's failing because it relies on code that only exists in the master branch of the compiler. We could get around this by putting a compiler commit SHA in the Travis environment, cloning and checking that out, and then using a sandbox with add-source? Alternatively we could try using stack, or waiting for the next compiler release.

paf31:
:+1: :+1:  :+1:  Done

PR #114, opened 2015-07-15
Fix error when searching for "on"

hdgarrood:
@beckyconning noticed that searching for "on" will cause an error: http://pursuit.purescript.org/search?q=on

This commit fixes that error.

---

Perhaps the "locations" thing lists all the places something is
exported, including re-exports.  The old code would throw an error
unless there was exactly one thing in the "locations" list. For now, it
seems to make the most sense to treat each one as a separate result
instead.

The actual data causing the error was one of the results for "on", which
had two values in the locations list: AssocNone from both string-parsers
and parsing.


paf31:
:+1: Looks good to merge. Thanks!

PR #120, opened 2015-07-17
Allow conditional caching, fix #117

### LINKED ISSUE #117: Package pages should not be cached if README fetching failed
Currently we cache packages even if we failed to obtain a README. Instead, I think we should only save a copy of the page on the disk if we successfully got it.

I think we can implement this by adding a variant of `cacheHtml` which expects you to return a tuple: `(OkToCache, Html)`, where `OkToCache` is isomorphic to `Bool`.

Also, once this is fixed and deployed, it's probably worth deleting all of the cached pages with no readme available. I think something like this would do it:

```
cd data/cache
grep -i -l "no readme available" > files-that-should-not-have-been-cached.txt
xargs rm < files-that-should-not-have-been-cached.txt
```

hdgarrood:
Add cacheConditional, which allows conditional caching of responses.
This allows us to avoid caching package pages which say "readme not
available" as a result of failing to retrieve the readme from github.

### DIFF
--- a/src/Handler/Caching.hs
+++ b/src/Handler/Caching.hs
@@ -1,7 +1,10 @@
 
 module Handler.Caching
-  ( cache
+  ( OkToCache(..)
+  , cache
+  , cacheConditional
   , cacheHtml
+  , cacheHtmlConditional
   , cacheSvg
   , cacheJSON
   , cacheText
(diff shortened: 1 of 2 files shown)


src/Handler/Caching.hs
```diff
@@ -23,13 +26,27 @@ import Data.Aeson (encode)
...
+
 -- | This function allows an upstream server (such as nginx) to cache a
 -- response, by writing the response body to a file before sending it to the
 -- client. Afterwards, the upstream server can respond to requests for the same
 -- path without having to proxy to the Yesod application, and any caching
 -- mechanisms that the upstream server supports can additionally be used.
-cache :: (a -> LB.ByteString) -> String -> Handler a -> Handler a
-cache toLbs basename action = action >>= (\body -> write body $> body)
+--
+-- We use Yesod's short-circuiting behaviour here to ensure that this code is
```
paf31: Where does the short-circuiting piece come in?

hdgarrood: If the `Handler` action that gets passed in uses `sendResponseStatus`, that short-circuits and sends a response immediately, and no further `Handler` code is run - in particular, this code shouldn't be reached.

paf31:
Thanks!

PR #121, opened 2015-07-17
Update LICENSE with instructions for relinking

hdgarrood:
... with modified versions of LGPL components, as per section 6a of the
LGPL v2.1

This should allow us to use Travis for building versions and publishing the results via GH releases.


hdgarrood:
Ended up doing a bit more here as well.

* Travis uploading releases should be working now (I'll test after this is merged)
* Speed builds up: used to be ~30 minutes, now ~6.

I think this is good to merge now.

paf31:
Looks great!

PR #131, opened 2015-07-19
Parameterize js

hdgarrood:
Rather than having the bulk of the JavaScript code in widget files, and
having rawJS scattered everywhere, we now extract as much code as
possible into the file static/js/Pursuit.js, and just have a single
function call in any widget file.

This way, we can reduce repeatedly transmitting JavaScript files which
only differ by a few bytes. It also feels cleaner.

---

We should also use nginx for serving static files, so that we can cache things (I think it will automatically take care of ETag, Last-Modified-Since, etc). I think something like this should do the job:

```
location /static {
  root /var/pursuit/static; # or whatever it is
}
```


paf31:
Sorry I missed this. Looks good!

PR #138, opened 2015-08-01
Distinguish types and values in links

### LINKED ISSUE #110: Fragments to type class members and data constructors don't work
For example, try clicking on the first result in http://new-pursuit.purescript.org/search?q=bind

One way of fixing this could be to make them point to the parent type class / data declaration instead. The only place where these links currently appear (as far as I'm aware) is in Hoogle results, and it's not currently obvious how to implement this. Perhaps it will be easier once we're using Hoogle 5. Alternatively, we could give each type class member / data constructor its own id attribute.

hdgarrood:
Fixes #110


hdgarrood:
On second thoughts, I think this is ok as it is, I don't think we need to fiddle with the CSS like I said before.

paf31:
:+1: Thanks!

PR #143, opened 2015-08-14
Add note about --dry-run flag in help page

### LINKED ISSUE #142: Update the help re new --dry-run flag

hdgarrood:
Fixes #142 

I also added paragraphs inside each `<li>`, which increases spacing between each item, making it easier to read. Without this, it would look very odd as the spacing between the `<pre>` and the paragrpah either side in the new item would be larger than the spacing between any two individual items.

### DIFF
--- a/templates/help.hamlet
+++ b/templates/help.hamlet
@@ -19,27 +19,41 @@
     <h3>How to submit packages
     <ol>
       <li>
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Looks good, thanks!

PR #145, opened 2015-08-14
Fix source links on windows, fixes #144

### LINKED ISSUE #144: Broken source links for packages produced on Windows
http://pursuit.purescript.org/packages/purescript-either/0.2.1/docs/Data.Either.Unsafe#d:fromRight

Not sure what's going on here.

hdgarrood:
The `SourceSpan` values in the data that `psc-publish` produces on
Windows are file paths separated by backslashes. This commit modifies
the code that generates the source links in order to account for this.

### DIFF
--- a/src/Model/DocsAsHtml.hs
+++ b/src/Model/DocsAsHtml.hs
@@ -203,9 +203,17 @@ linkToSource (LinksContext{..}, _) (P.SourceSpan name start end) =
   (P.SourcePos endLine _) = end
   (GithubUser user, GithubRepo repo) = ctxGithub
 
-  relativeToBase = intercalate "/" . dropWhile (/= "src") . splitOn "/"
+  relativeToBase = intercalate "/" . dropWhile (/= "src") . splitOnPathSep
   githubBaseUrl = concat ["https://github.com/", user, "/", repo]
   fragment = "L" ++ show startLine ++ "-L" ++ show endLine
+
+-- | Split a string on either unix-style "/" or Windows-style "\\" path
+-- | separators.
+splitOnPathSep :: String -> [String]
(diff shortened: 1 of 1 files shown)


src/Model/DocsAsHtml.hs
```diff
@@ -203,10 +203,18 @@ linkToSource (LinksContext{..}, _) (P.SourceSpan name start end) =
...
   (GithubUser user, GithubRepo repo) = ctxGithub
 
-  relativeToBase = intercalate "/" . dropWhile (/= "src") . splitOn "/"
+  relativeToBase = intercalate "/" . dropWhile (/= "src") . splitOnPathSep
   githubBaseUrl = concat ["https://github.com/", user, "/", repo]
   fragment = "L" ++ show startLine ++ "-L" ++ show endLine
 
+-- | Split a string on either unix-style "/" or Windows-style "\\" path
+-- | separators.
+splitOnPathSep :: String -> [String]
```
paf31: Why not use a regex to split on either character?

paf31:
:ship:

src/Model/DocsAsHtml.hs
hdgarrood: I think regexes don't belong in real code; in most cases they're near write-only and very difficult to debug. The one you describe would be fairly straightforward, but I can't see an improvement over what's currently here.

Also, I don't know how to split on a regex in Haskell, because of the above :)

paf31:
I guess something like `Data.List.Split` would be a better match than a regex in this case anyway, since you can say "split on any of the things in this list".

paf31:
Looks good as is though.

PR #151, opened 2015-09-08
Package index by first letter

### LINKED ISSUE #146: Group package list by first letter
And possibly show names of packages in two columns, if there is space?

paf31:
Fixes #146.

I'm testing this using Travis since I can't get things to build easily right now. I expect it'll fail since I've never used Hamlet before.

### DIFF
--- a/templates/homepage.hamlet
+++ b/templates/homepage.hamlet
@@ -16,7 +16,8 @@
     <h2>No packages uploaded yet.
   $else
     <h2>Packages:
-    <ul>
-      $forall pkgName <- pkgNames
-        <li>
+    $forall group <- groupBy ((==) `on` (toUpper . head . runPackageName)) pkgNames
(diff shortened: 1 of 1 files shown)


hdgarrood:
:+1: lgtm

paf31:
Hmm, maybe I should have put this into https://github.com/purescript/pursuit/blob/aad2298a09c70ff2b4c0f0c3f9f5da103cee5ce9/src/Handler/Packages.hs#L19 now that I look more?

hdgarrood:
Possibly. You could also argue that the current formulation is better: being "view" code, it should go into the "view" (rather than, "model" or "controller"). I'm not sure.

From a less idealistic perspective, putting it into a normal Haskell file will make error messages easier to read, if, say, someone wants to change that code and introduces a type error.

Issue #156, opened 2015-11-04
Version numbers are not sorted properly


tfausak:
The version select box displays 0.0.10 below 0.0.2 and considers 0.0.9 the latest version.

![screen shot 2015-11-04 at 7 08 48 am](https://cloud.githubusercontent.com/assets/408753/10938894/475a2f02-82c3-11e5-91de-218c5facf3ee.png)

http://pursuit.purescript.org/packages/purescript-neon/0.0.10

However if I go to <http://pursuit.purescript.org/packages/purescript-neon>, it takes me to 0.0.10.

hdgarrood:
Such type safety. Wow.

*hangs head in shame*

tfausak:
To be fair, I think the problem is in JavaScript: https://github.com/purescript/pursuit/blob/aad2298/static/js/Pursuit.js#L43

hdgarrood:
Yes, but there's a little more to it than that I think. The versions should have been sorted while they were still values of type `Version` on the server side, rather than `String`; that is, before being serialized. I already fixed it, see #157 :)

tfausak:
:+1:

PR #157, opened 2015-11-04
Fix incorrect version sorting - #156

hdgarrood:
Sort versions *before* `show`ing them, so that they get sorted as
versions, not as strings.


paf31:
Looks good, thanks! I'll deploy this soon.

paf31:
Deployed, but still seems to be happening. Not sure if I need to do anything except update the executable.

hdgarrood:
`rm data/cache/**/*.json` I think

hdgarrood:
(although probably look at a list of those files first, to be safe)

paf31:
Yep, sorry, didn't update the `static` directory either.

paf31:
Looks good now :+1: Thanks!

Issue #162, opened 2015-11-17
Badges aren't being updated


zudov:
For example http://pursuit.purescript.org/packages/purescript-strongcheck-generics/badge/ says `0.2.0` while it should be `0.2.1` (updated about a day ago).

Another example is http://pursuit.purescript.org/packages/purescript-websocket-simple/badge which is 404 even though the package is present (I've uploaded it a few minutes ago).

I guess there is a cache in front of badge endpoint, perhaps it should be adjusted.
I guess 404 appeared because I've made a few requests to `purescript-websocket-simple`'s badge endpoint *before* I published it, so 404 got cached.

hdgarrood:
I know why the out of date badge is happening: I forgot to clear the cached badge when new package versions are uploaded.

The 404 is weird. That might be because of the change in #160? @paf31 could you check what happens if you request `GET /packages/purescript-websocket-simple/badge` directly to the Yesod server without going via nginx?

paf31:
`v0.0.1`

hdgarrood:
And now it appears to be working from here, presumably because the SVG has been written to disk and it's being served directly by nginx from the disk. I am becoming more convinced that the problem is in the nginx configuration. If you delete the cache for that badge, and then request the same badge again from outside (ie through nginx), can you check if the request reaches the Yesod server?

paf31:
The server only has `0.0.1.json`. Removing the cache and `curl`ing again still gives the badge for `0.0.1`.

paf31:
Also `index.html` wasn't getting updated, and I had to delete it to show the new items on the front page. Not sure if that's Nginx or Yesod though.

hdgarrood:
oh as in the root of the whole site isn't getting updated? Oh yeah, that may well be another one I forgot to clear.

Um, if requesting via nginx after deleting the cache is working now then I don't know what was going on. I would have thought that would put us back in the original state where we were getting 404s.

hdgarrood:
Closing for now as the stale badge issue has been resolved and we're unable to reproduce the 404 issue. If it happens again then I'll have another look.

zudov:
http://pursuit.purescript.org/packages/purescript-strongcheck-generics/badge/ is still `v0.2.0`, even though the latest one is `v0.3.0`. Would new release force the update?

hdgarrood:
The fix is in master but it hasn't been deployed yet.

zudov:
@hdgarrood Ok. Thanks!

PR #163, opened 2015-11-17
Fix badge cache invalidation (refs #162)

hdgarrood:
Ensure cached versions of package badges are invalidated properly. Partial fix for #162.

### DIFF
--- a/src/Handler/Caching.hs
+++ b/src/Handler/Caching.hs
@@ -82,25 +82,46 @@ cacheText = cache LTE.encodeUtf8 "index.txt"
 
 -- | Clear the whole cache for a particular package at a particular version.
 -- Called whenever a new version of a package is uploaded.
+--
+-- Note that this function is also responsible for clearing cached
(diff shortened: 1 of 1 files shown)


paf31:
Why is this only a partial fix? What remains to be fixed?

hdgarrood:
The 404 for http://pursuit.purescript.org/packages/purescript-websocket-simple/badge is still unexplained.

paf31:
I don't get a 404 there any more. Was this before I cleared the cache?

hdgarrood:
Um, I'm not sure. What does "this" mean in this context?

paf31:
The 404. I'm a bit lost here, is there still an issue that we know of?

hdgarrood:
Yeah I don't know. I guess you could try reproducing it by deleting the entire cache for purescript-websocket-simple and trying to GET the badge again. If that doesn't reproduce it then maybe we can just give up for now and hope it doesn't happen again.

PR #181, opened 2016-02-02
Updates for purescript 0.8

hdgarrood:


hdgarrood:
Seems to work, so I'm going to merge. The hard bit is dealing with the existing data.

PR #184, opened 2016-02-02
Distinguish package uploader and owner in UI

hdgarrood:
Makes the UI a bit less confusing. The uploader is not particularly
important info to most people browsing Pursuit, so we move that info to
the bottom. Also switch "published" for the less ambiguous "uploaded".


paf31:
Looks good, thanks!

PR #187, opened 2016-02-04
Fix Hoogle database generation errors

hdgarrood:
They were arising from the use of the Show ModuleName instance, which
previously was equivalent to the current runModuleName, but now is a
real Show instance.

Locally, using the most recent backup, this brings the number of Hoogle database regen errors from ~7000 down to 179. :smile:


paf31:
Deployed as part of v0.3.2.

PR #190, opened 2016-02-29
Add "/" as a global search-input focus shortcut

spicydonuts:


hdgarrood:
Thanks for the contribution, but I'm not very keen on this, given that Firefox already uses that key for a quick search feature.

hdgarrood:
(That's why I chose S in the first place, even.)

spicydonuts:
Ah, I didn't know that.  Like Ctrl+F search?  Just used to "/" in most tools : ] (vim, less, twitter).

hdgarrood:
Yeah, it's similar to that. Not sure why they needed two but there you go

spicydonuts:
(just discovered npmjs.org uses "/" too : ]  )

michaelficarra:
Yeah, a lot of websites use `/`. But as a counter-example, GitHub uses `s`.

Issue #166, opened 2015-11-21
Fragment encoding issue with search results


hdgarrood:
Repro steps:
* Search http://pursuit.purescript.org/search?q=footer%27
* Click on the first result

The fragment you get is `#v:footer-39-` but it should really be `#v:footer'` (or maybe we should percent-encode the `'`, I'm not quite sure).

kseo:
#193 fixed this. Now we get the fragment `#v:footer'` correctly.

hdgarrood:
Confirmed, thank you!

PR #193, opened 2016-03-14
Fix a typo.

kseo:
(>>) is accidentally used where (>>>) is required. This small typo causes the
(decodeAnchorId >>> bracketOperators) to be skipped.


hdgarrood:
Thanks! Could you please expand on the exact problem that this causes and how it manifests from a  user's perspective?

kseo:
#166 is caused by this typo. You get `#v:footer-39-` instead of `#v:footer'` because `decodeAnchorId` is skipped.

hdgarrood:
Ah, excellent - thank you!

PR #204, opened 2016-03-28
README: More info about server start up

LiamGoodacre:
Some details which I would have found useful.

Is the wording okay?  Under the right section?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ To run the server:
 $ stack exec pursuit
 ```
 
+If you have content in your data directory
+(see [Database structure](#database-structure)) then the Hoogle database will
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think I would replace "package database" with "Hoogle database", as, in my mind, the "package database" just refers to all the files under `data/verified` (which do not get modified or 'regenerated' in any sense during this operation). Other than that, :+1: 

Also re your IRC comment, I would expect the first `stack build` to take a long time; there are lots of dependencies. Thankfully, that long wait is a one-time thing.

hdgarrood:
Super, thanks! :)

LiamGoodacre:
:smile:

Issue #185, opened 2016-02-03
Keep query in search field


AppShipIt:
As a user I'd like to edit my previously entered query in the search field after navigating to results 🙏

hdgarrood:
:+1: as you pointed out on twitter, this could be useful for if you make a typo

LiamGoodacre:
Hey, I'm having a go at this :smiley:
PR coming shortly.

hdgarrood:
Fixed by #185. Thanks again @LiamGoodacre!

PR #205, opened 2016-03-28
For #185 "Keep query in search field"

LiamGoodacre:


src/Foundation.hs
```diff
@@ -106,6 +106,7 @@ instance Yesod App where
 
         manalytics <- appAnalytics . appSettings <$> getYesod
         isSearch <- testCurrentRoute (== SearchR)
+        searchText <- map (maybe "" unpack) (lookupGetParam "q")
```
hdgarrood: Would `fromMaybe ""` work here? I think the templates ought to work the same way when splicing `Text` values in as opposed to `String` values.

LiamGoodacre: Yep, good call.

hdgarrood:
:+1: thank you!

LiamGoodacre:
Thanks!

PR #211, opened 2016-05-01
Use barrier to generate badges

### LINKED ISSUE #102: Use shields.io for badges
I can't quite work out what it is, but the badges that pursuit currently renders just don't quite look right. Eg: http://new-pursuit.purescript.org/packages/purescript-thermite/0.8.0

tfausak:
This fixes #102.

I decided to use barrier because it's what Stackage uses. Check out [their badge handler](https://github.com/fpco/stackage-server/blob/ade312b9dfa98db27f2eca68c15b75036ce128c1/Handler/Package.hs#L53) to see how it works.

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -123,6 +123,7 @@ library
                  , cookie
                  , xss-sanitize
                  , wai-middleware-gunzip
+                 , barrier ==0.1.*
 
     if flag(dev)
       build-depends: foreign-store
--- a/src/Handler/PackageBadges.hs
+++ b/src/Handler/PackageBadges.hs
@@ -3,11 +3,10 @@ module Handler.PackageBadges where
 
 import Import
 import Data.Version
-import Text.Blaze (toValue)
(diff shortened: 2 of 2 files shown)


tfausak:
Almost forgot to post screenshots! Before:

screen shot 2016-05-01 at 6 54 25 am

After:

screen shot 2016-05-01 at 6 52 53 am

hdgarrood:
Looks great :)

Just one thing: I think I'd rather use the `renderBadge` function, since then I don't think we need to import the `Internal` module. Then I think you can use `preEscapedToMarkup` to turn the `ByteString` into `Svg`.

tfausak:
👍 I had to use [`unsafeLazyByteString`](http://haddock.stackage.org/lts-5.14/blaze-markup-0.7.0.3/Text-Blaze.html#v:unsafeLazyByteString) instead of `preEscapedToMarkup` because there's no `ByteString` instance for `ToMarkup`.

hdgarrood:
Looks great, thanks!

Issue #100, opened 2015-07-05
Use semver instead of Data.Version


hdgarrood:
Prerequisite for #98, probably

semver is MPL, but I don't think licensing is an issue, see https://www.mozilla.org/MPL/2.0/FAQ.html#distribute-my-binaries and https://www.mozilla.org/MPL/2.0/FAQ.html#virality.

hdgarrood:
Most of the work required for this will have to be inside the compiler, in `Language.PureScript.Docs.Types`. I'd like to achieve this without breaking the JSON format; the way to do this would be to serialize versions as strings in the normal way (eg, "1.2.3") before putting them in, as we do currently.

It's possibly worth noting that this would break existing JSON documents produced by `psc-publish` where the version is in four components, which `psc-publish` currently doesn't stop you from doing. However, this is very rare for PureScript packages because we use Bower (which does require semver).

PR #219, opened 2016-06-15
Update for PureScript 0.9.1

jplatte:
I've tested this with `purescript-prelude`, `purescript-eff`, `purescript-console` and `purescript-psci-support` in `data/verified`. The only user-facing change is that operator fixity is not included in the text below an operator definition anymore, because it is now included in the definition itself. We could potentially do better at rendering operator aliases, but I'd say this isn't all bad either:
bildschirmfoto am 2016-06-15 um 18 03 03

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -85,7 +85,7 @@ library
                  , directory                     >= 1.1        && < 1.3
                  , warp                          >= 3.0        && < 4
                  , data-default
-                 , aeson                         >= 0.6        && < 0.10
+                 , aeson                         >= 0.11       && < 0.12
                  , conduit                       >= 1.0        && < 2.0
                  , monad-logger                  >= 0.3        && < 0.4
                  , fast-logger                   >= 2.2        && < 2.5
@@ -96,7 +96,7 @@ library
                  , containers
                  , vector
                  , time
-                 , purescript >= 0.8.0
+                 , purescript >= 0.9.1
                  , bower-json >= 0.4
                  , lucid
                  , blaze-builder
--- a/src/Application.hs
+++ b/src/Application.hs
@@ -17,7 +17,7 @@ import "monad-logger" Control.Monad.Logger
   (liftLoc)
 import Language.Haskell.TH.Syntax
   (qLocation)
-import Control.Concurrent (threadDelay, forkIO)
+import Control.Concurrent (forkIO)
 import Network.Wai.Handler.Warp
   (Settings, defaultSettings, defaultShouldDisplayException, runSettings,
   setHost, setOnException, setPort, getPort)
--- a/src/Model/DocsAsHtml.hs
+++ b/src/Model/DocsAsHtml.hs
@@ -18,7 +18,7 @@ module Model.DocsAsHtml (
 import Prelude
 import Control.Arrow (second)
 import Control.Category ((>>>))
-import Control.Monad (when)
+import Control.Monad (unless)
 import Data.Char (toUpper)
 import Data.Ord (comparing)
 import Data.Monoid ((<>))
@@ -128,22 +128,26 @@ declAsHtml r ctx d@Declaration{..} = do
     linkTo declFragment $
       h3_ (text declTitle)
     div_ [class_ "decl-inner"] $ do
-      code_ [class_ "code-block"] $
-        codeAsHtml r ctx (Render.renderDeclaration d)
+      case declInfo of
+        AliasDeclaration fixity alias ->
+          renderAlias fixity alias
+        _ ->
+          code_ [class_ "code-block"] $
+            codeAsHtml r ctx (Render.renderDeclaration d)
 
       for_ declComments renderComments
 
       let (instances, dctors, members) = partitionChildren declChildren
 
-      when (not (null dctors)) $ do
+      unless (null dctors) $ do
         h4_ "Constructors"
         renderChildren r ctx dctors
 
-      when (not (null members)) $ do
+      unless (null members) $ do
         h4_ "Members"
         renderChildren r ctx members
 
-      when (not (null instances)) $ do
+      unless (null instances) $ do
         h4_ "Instances"
         renderChildren r ctx instances
 
@@ -239,6 +243,21 @@ splitOnPathSep str
   | '/'  `elem` str = splitOn "/" str
   | '\\' `elem` str = splitOn "\\" str
   | otherwise       = [str]
+
+renderAlias :: P.Fixity -> FixityAlias -> Html ()
+renderAlias (P.Fixity associativity precedence) alias =
+  p_ $ do
+    text $ "Operator alias for " <> P.showQualified showAliasName alias <> " "
+    em_ (text ("(" <> associativityStr <> " / precedence " <> show precedence <> ")"))
+  where
+  showAliasName (Left valueAlias) = P.runProperName valueAlias
+  showAliasName (Right typeAlias) = case typeAlias of
+    (Left identifier)  -> P.runIdent identifier
+    (Right properName) -> P.runProperName properName
+  associativityStr = case associativity of
+    P.Infixl -> "left-associative"
+    P.Infixr -> "right-associative"
+    P.Infix  -> "non-associative"
 
 -- TODO: use GitHub API instead?
 renderComments :: String -> Html ()
--- a/src/Model/DocsAsHtml.hs
+++ b/src/Model/DocsAsHtml.hs
@@ -131,7 +131,6 @@ declAsHtml r ctx d@Declaration{..} = do
       code_ [class_ "code-block"] $
         codeAsHtml r ctx (Render.renderDeclaration d)
 
-      for_ declFixity renderFixity
       for_ declComments renderComments
 
       let (instances, dctors, members) = partitionChildren declChildren
@@ -240,15 +239,6 @@ splitOnPathSep str
   | '/'  `elem` str = splitOn "/" str
   | '\\' `elem` str = splitOn "\\" str
   | otherwise       = [str]
-
-renderFixity :: P.Fixity -> Html ()
-renderFixity (P.Fixity associativity precedence) =
-  p_ (em_ (text (associativityStr <> " / precedence " <> show precedence)))
-  where
-  associativityStr = case associativity of
-    P.Infixl -> "left-associative"
-    P.Infixr -> "right-associative"
-    P.Infix  -> "non-associative"
 
 -- TODO: use GitHub API instead?
 renderComments :: String -> Html ()
--- a/src/Import/NoFoundation.hs
+++ b/src/Import/NoFoundation.hs
@@ -3,15 +3,10 @@ module Import.NoFoundation
     , module Import
     ) where
 
-import ClassyPrelude.Yesod   as Import hiding (intercalate)
+import ClassyPrelude.Yesod   as Import
 import Settings              as Import
 import Yesod.Core.Types      as Import (loggerSet)
 import Yesod.Default.Config2 as Import
 import Control.Category      as Import ((>>>), (<<<))
 
 type One = Succ Zero
-
--- This is a more useful way of defining intercalate, imo.
-intercalate :: (Monoid (Element seq), SemiSequence seq) =>
-  Element seq -> seq -> Element seq
-intercalate x xs = concat (intersperse x xs)
--- a/src/Model/DocsAsHoogle.hs
+++ b/src/Model/DocsAsHoogle.hs
@@ -10,7 +10,7 @@ module Model.DocsAsHoogle
 -- | an input file, suitable for supplying to Hoogle to build a database from.
 
 import Prelude
-import Control.Arrow (first, second)
+import Control.Arrow (first)
 import Data.Maybe (fromMaybe)
 import Data.List (intersperse)
 import Data.Monoid
@@ -162,7 +162,9 @@ renderChildDeclaration
         classConstraint =
           case D.declInfo parent of
             D.TypeClassDeclaration args _ ->
-              (P.Qualified Nothing (P.ProperName (D.declTitle parent)), map D.toTypeVar args)
+               P.Constraint { constraintClass = P.Qualified Nothing (P.ProperName (D.declTitle parent))
+                            , constraintArgs = map D.toTypeVar args
+                            , constraintData = Nothing }
             _ ->
               invalidArgument $
                 "the parent of a type class member was something other than a "
@@ -226,7 +228,7 @@ everywhereOnTypesInfo f info =
   case info of
     D.ValueDeclaration ty -> D.ValueDeclaration (go ty)
     D.TypeSynonymDeclaration args ty -> D.TypeSynonymDeclaration args (go ty)
-    D.TypeClassDeclaration args implies -> D.TypeClassDeclaration args (map (second (map go)) implies)
+    D.TypeClassDeclaration args implies -> D.TypeClassDeclaration args (P.mapConstraintArgs (map go) <$> implies)
     other -> other
   where
   go = P.everywhereOnTypesTopDown f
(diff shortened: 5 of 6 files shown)


pursuit.cabal
```diff
@@ -85,7 +85,7 @@ library
                  , directory                     >= 1.1        && < 1.3
                  , warp                          >= 3.0        && < 4
                  , data-default
-                 , aeson                         >= 0.6        && < 0.10
+                 , aeson                         >= 0.6        && < 0.12
```
hdgarrood: Since we are using aeson 0.11.x, I'd prefer to bump up the lower bound to 0.11, so that we can't break the format just by choosing a different version of aeson.

src/Application.hs
```diff
@@ -12,7 +12,8 @@ module Application
     , handler
     ) where
 
-import Import
+-- TODO: How is threadDelay even exported from there? Should that be changed instead of hiding it here?
+import Import hiding (threadDelay)
```
hdgarrood: Is this a different `threadDelay` to the one exported by `Control.Concurrent`?

pursuit.cabal
jplatte: Wait, what? Does `< 0.11` match `0.11.x`??

src/Application.hs
jplatte: Apparently so. Here is the error message when you don't hide the one from `Import`:

    Ambiguous occurrence ‘threadDelay’
    It could refer to either ‘Import.threadDelay’,
                             imported from ‘Import’ at src/Application.hs:16:1-13
                             (and originally defined in ‘lifted-base-0.2.3.6:Control.Concurrent.Lifted’)
                          or ‘Control.Concurrent.threadDelay’,
                             imported from ‘Control.Concurrent’ at src/Application.hs:21:28-38
                             (and originally defined in ‘GHC.Conc.IO’)

pursuit.cabal
hdgarrood: The `ToJSON` and `FromJSON` instances that Aeson provides for common data types might have changed between versions 0.x and 0.(x+1), and the parsing of the JSON files that people upload makes use of those instances. We need to be using the same version as what the compiler is using, essentially. Really the bounds should have been `>= 0.9 && < 0.10` before, I think.

hdgarrood: I mean I don't want to worry about everything breaking in case someone uses `cabal-install` instead of `stack` for whatever reason, and the dependency solver selects an older version of `aeson`.

jplatte: Okay, now I know what you mean. Does it affect the running Travis build when I overwrite this commit / push another one? (seeing as that takes quite a while and we don't really need to run it another time for this change because it already picks aeson 0.11.2.0)

hdgarrood: It will kick off a new Travis build if you do that, yeah. You can add `[ci skip]` to the commit message to stop Travis from doing a CI build, I think.

src/Application.hs
hdgarrood: Ah, I see. Is the version from `Import` a generalised version of the one from `Control.Concurrent` which could also work in this module? If so, then we could just remove `threadDelay` from the `Control.Concurrent` import list. If not, I think the code you've got here is fine.

I would guess the one from Import is coming via Import.NoFoundation and from there, via ClassyPrelude.Yesod: https://github.com/purescript/pursuit/blob/e1dc3550ff88473bdfba003820743ddac0bcf688/src/Import/NoFoundation.hs#L6

hdgarrood:
Oh, that reminds me, I think we can get rid of the `intercalate` defined in `Import.NoFoundation` now, since I sent a PR to classy-prelude addressing the issue that lead me to do that in the first place: https://github.com/snoyberg/classy-prelude/pull/119. Now that we're on LTS 6 I think we will be getting a version of classy-prelude that includes that change.

jplatte:
Alright. I had a look at `threadDelay` from `lifted-base`, seems like it is a generalized version. So I guess it makes the most sense to do all the changes in one rewrite of my commit now, and restart the Travis build.

hdgarrood:
👍

jplatte:
Just found something problematic... hoogle doesn't seem to be able to parse all of the data we feed it:

```
Parse error 5:2: Unknown attribute: version
Parse error 11:17: Parse error: !
Parse error 65:2: Unknown attribute: version
Parse error 75:13: Parse error: #
Parse error 171:2: Unknown attribute: version
Parse error 254:10: Parse error: Control.Apply.apply
Parse error 263:10: Parse error: Control.Apply.applyFirst
Parse error 272:10: Parse error: Control.Apply.applySecond
Parse error 331:10: Parse error: Control.Bind.bind
Parse error 342:10: Parse error: Control.Bind.bindFlipped
Parse error 361:10: Parse error: Control.Bind.composeKleisli
Parse error 370:10: Parse error: Control.Bind.composeKleisliFlipped
Parse error 464:10: Parse error: Control.Semigroupoid.compose
Parse error 473:10: Parse error: Control.Semigroupoid.composeFlipped
Parse error 584:9: Parse error: Data.Eq.eq
Parse error 594:9: Parse error: Data.Eq.notEq
Parse error 624:10: Parse error: Data.EuclideanRing.div
Parse error 677:10: Parse error: Data.Function.apply
Parse error 698:10: Parse error: Data.Function.applyFlipped
Parse error 733:10: Parse error: Data.Functor.map
Parse error 744:10: Parse error: Data.Functor.mapFlipped
Parse error 767:10: Parse error: Data.Functor.voidRight
Parse error 776:10: Parse error: Data.Functor.voidLeft
Parse error 844:10: Parse error: Data.HeytingAlgebra.conj
Parse error 848:10: Parse error: Data.HeytingAlgebra.disj
Parse error 869:10: Parse error: type
Parse error 927:10: Parse error: Data.Ord.lessThan
Parse error 931:10: Parse error: Data.Ord.lessThanOrEq
Parse error 935:10: Parse error: Data.Ord.greaterThan
Parse error 939:10: Parse error: Data.Ord.greaterThanOrEq
Parse error 1051:10: Parse error: Data.Ring.sub
Parse error 1085:10: Parse error: Data.Semigroup.append
Parse error 1131:10: Parse error: Data.Semiring.add
Parse error 1135:10: Parse error: Data.Semiring.mul
Parse error 1202:2: Unknown attribute: version
```

It seems like the hoogle converter doesn't handle operator aliases properly yet (searching for any operator gives me 0 results). Not sure if that should be counted as a 0.9.1 changes, as they were available on 0.8.x already, but I think they weren't used that much (not even in Prelude). I'll just upload the latest changes for now. If I'm going to fix this hoogle problem, that's at least going to be in a seperate commit.

hdgarrood:
Ah, yes. I would like to fix operator aliases properly before we update Pursuit, I think. The "Unknown attribute: version" is not new and I don't think it's worth worrying about.

src/Model/DocsAsHtml.hs
```diff
@@ -241,15 +240,6 @@ splitOnPathSep str
   | '\\' `elem` str = splitOn "\\" str
   | otherwise       = [str]
 
-renderFixity :: P.Fixity -> Html ()
-renderFixity (P.Fixity associativity precedence) =
```
hdgarrood: I think we should be able to leave this in - the information is still there, as far as I know, it's just been moved into the `DeclarationInfo`.

hdgarrood:
Can you have a quick look to see how difficult it would be to link to the relevant data constructor / type / value in the docs for an operator alias?

src/Model/DocsAsHtml.hs
jplatte: Yes, I think we're able to. But what point is there in rendering it as another paragraph when the same information is in the declaration already?

hdgarrood: Because it's easier to understand? I don't really like the `<code>` part of the rendering of fixity declarations actually, if we're getting rid of one of them I'd rather it were that one.

jplatte:
What about the row stuff, though (on the top)? Shouldn't that be converted already? I had a look at the actual data that results in these errors, and what it complains about are the definitions of `CONSOLE` and `Eff` (so as expected they won't show up in search results either). Also, I checked what the `type` error was for, and it is this line:

    infixr 4 type Data.NaturalTransformation.NaturalTransformation as ype (~>

(a data constructor alias). Really a strange but to have an extraneous thing like `ype (`.

hdgarrood:
> What about the row stuff, though (on the top)? Shouldn't that be converted already?

Sorry, I don't understand what you're referring to.

jplatte:
I meant the PureScript-specific kinds hoogle complained about at the beginning of the error log:

    Parse error 11:17: Parse error: !
    Parse error 75:13: Parse error: #

hdgarrood:
Ah, right. No, I don't think we attempt to do anything about those right now. If we're going to try to fix that, I'd rather do it properly, ie by improving support for languages other than Haskell in Hoogle. Last I heard this was one of the goals for Hoogle 5.

jplatte:
Okay. So I'll first look at replacing the operator code block with text (including a link to the function), as the HTML stuff seems simpler than Hoogle.

jplatte:
So this is how it looks now:

bildschirmfoto am 2016-06-18 um 14 58 09
bildschirmfoto am 2016-06-18 um 14 57 42

Not sure if the `type` in front of a type alias is meant to be there, but I didn't do that. The other thing is that the names are qualified even though I'd say they don't need to be. That was already the case with the code blocks and I think is because the purescript library has this 'current module' thing somewhere that we probably don't set properly, right? I think it would be good if we could fix that, but I guess for now it works too.

Also, I'd be interested in knowing whether some of the stuff I wrote could have been written more easily, because I don't like the look of what I wrote that much.

I've had a look at the hoogle stuff as well, but I really don't know where to start – the `codeAsHoogle` function seems like the right starting point, but that operates on already rendered code, and it seems unnecessarily hard to transform a rendered alias declaration into a haskell-style operator definition, as opposed to rendering the alias declaration as that.

paf31:
@hdgarrood When you're happy with this, I can deploy to the server. I assume we'll need to do another backup-delete-repopulate run using your script too?

jplatte:
@paf31 We'll need to fix the operator aliases we give to Hoogle first. Currently, searching for operators will always give you no results. And like I said I don't know where to start with changing the Hoogle "conversion" (or whatever you'd call it).

jplatte:
Bump! @hdgarrood

paf31:
In the long term, I think it might be worth thinking about replacing type search with an in-memory database using the `subsumes` function from the typechecker. Hoogle 5 has been in the works for a while, with some unknowns around support for other languages.

Name search could use something like ElasticSearch, which would give us the benefit of enabling full text search in documentation.

The full Pursuit JSON database currently takes up about 650M on disk, and that's JSON. Compressed in memory, it could take up far less.

---

There's no immediate rush to get this released. I suspect @hdgarrood is busy at the moment.

hdgarrood:
Yeah, sorry, I am going to get around to looking at this properly soon.

hdgarrood:
I'm a little apprehensive to use the compiler in that way for type search, because I think we'd just be duplicating effort. I think I'd rather put that effort into getting Hoogle 5 out. Also, the more compiler internals we depend on, the more often we have to break and re-upload everything. Unless we start deferring all the processing to the server-side, I guess? Have we considered uploading source distributions to pursuit; i.e., doing the processing after the uploading instead of before?

paf31:
> I think I'd rather put that effort into getting Hoogle 5 out.

Ultimately, it's your call, but I think we can come up with something lightweight which is tailored to our specific type system. I've not been keeping up with Hoogle 5's development, but it seems like it could be potentially quite tricky to support all the type system features we want in a generic way.

> Unless we start deferring all the processing to the server-side, I guess? Have we considered uploading source distributions to pursuit; i.e., doing the processing after the uploading instead of before?

I would be fine with that option. Didn't we discuss something similar recently.

hdgarrood:
Fair enough. I think either approach is quite a lot of work, and I'm not going be able to find the time to do either any time soon, so I guess it makes most sense to leave it up to whoever actually takes it on.

I don't remember discussing uploading purescript source instead of the current JSON recently; do you know if there's a link or anything?

hdgarrood:
@jplatte I think the problem with a) operator aliases in Hoogle, and b) the `type` appearing for type operator aliases is that we've lost too much information in the processing that `psc-publish` does, and I think the answer to this should probably be to upload PureScript source code instead of the current JSON data. This would also fix the problem of having to regenerate everything too often; by uploading source code, we should only need to regenerate packages when there are breaking changes to the language, and no more often than that. Here's a vague sketch of what that might look like.

* I think it would be good to have `psc-publish` do everything it does right now, but just discard most of the resulting information which previously would have been uploaded, and instead create an archive including the source code instead. The archive would also want to include all the selected versions of dependencies, so that the server has a working install plan and therefore doesn't need to any dependency solving to determine which dependencies to use when generating, and also use this information when linking to types in other packages.
* On upload, the server would attempt to do what `psc-publish` does now to produce the JSON data, and save the resulting JSON to disk (I expect this is a costly enough operation to be worth saving the JSON so that we only have to do this once). We could also include the compiler version in the file name of the JSON which we save to disk, so that upgrading the version of `purescript` that pursuit depends on doesn't require manual intervention in the DB (ie, deleting all the existing JSON files). I expect that this will be a long enough process that we will want to return a `202 Accepted` response immediately, to avoid risking connection timeouts or similar issues.
* It would be nice to have some kind of API to ask what state a recently uploaded package is in (processing, processed successfully, failed) so that tools like `pulp publish` can wait until a package has either been uploaded successfully or has failed, and then report this to the user. I guess we could put that API at `GET /packages/purescript-prelude/0.1.0/status` and have it return a small JSON object like `{"status": "success"}` or `{"status":"failed","error":"Dependency purescript-prelude 1.0.1 does not exist in the database"}`

I've literally just written this out without a huge amount of thought, so nothing here is set in stone. If something seems off, it probably is, so please do ask.

One thing that's just occurred to me is that if we changed Pursuit as above, it would mean that uploading a package that depends on some other package would now require the other package to exist on Pursuit at the exact same version that the author has installed locally (which is not currently the case). This doesn't seem like too bad a thing to me, though.

The first bullet point is going to require changes inside the compiler (it would be lovely if we could switch the representation of versions to a real semver library at the same time - see #100 - but this is already quite a big undertaking anyway, so perhaps not).

@paf31 How much hard disk space do we have? Do you think we'll have to start worrying about hard disk space with the above proposal?

jplatte:
@hdgarrood I honestly don't know anything about `psc-publish` so I can't comment on your proposal. I'm also pretty sure that with my current haskell proficiency / experience I wouldn't be able to implement this on my own, and unless someone wants to guide me along, I think I'm out.

What will happen with the current patch though? Shouldn't we at least deploy the docs for all the 1.0 libraries? Redeploying pursuit might also fix the search that is completely broken at the moment, at least to some extent.

In any case, I think you should create a new issue for the pursuit redesign because this PR doesn't really have anything to do with that.

hdgarrood:
The way it relates to this PR is that it's how I think we should fix the outstanding issues you mentioned in your recent comment: https://github.com/purescript/pursuit/pull/219#issuecomment-226941593

As you note, it may be worth deploying what we have here anyway, yes. I would like to find out what actually happened to the search instead of just redeploying and hoping the issue goes away, though.

paf31:
Bump. I know everyone's busy, I just want to make sure this one doesn't get forgotten.

Does it make sense to deploy a version without search, populate data for 0.9.1, and then add it back later?

hdgarrood:
I think that would make sense, yes. Given that the proposal I gave is probably not going to be implemented any time soon, I guess it makes most sense to deploy what we have now, so at least people using 0.9.1 can access documentation.

paf31:
I can merge this and deploy then, if you both like?
 
@hdgarrood If you get a spare half an hour, maybe we can sync up on a hangout and make an initial pass at diagnosing the search issue too.

hdgarrood:
Happy for this to be merged and deployed, but unfortunately I won't be able to set aside any time to help until next weekend.

paf31:
Well I can deploy, but I might need a hand repopulating the data. I can wait until next weekend, or if you can send me your script, I can have a go at running it.

hdgarrood:
The script is https://github.com/hdgarrood/pursuit-check-missing (although I usually end up doing a small amount of ad-hoc bash stuff too, which I didn't save). I've used that script for a few slightly different purposes as well, so not all the code in there will be relevant to this.

I guess the process will want to be slightly different this time, actually, as I think the set of package versions up now and the set that we will want to have up after updating Pursuit are entirely disjoint (for previous pursuit updates, this was not the case). Up to you whether you want to have a go now or wait until next weekend, I'm happy either way.

jplatte:
I agree, deploying this in its current state is probably the best way forward for now.

paf31:
@hdgarrood Do you have any time today to help me through the deployment? (no worries if not, I can do next week)

paf31:
I'm going to merge this and tag a release anyway. I'll try to deploy the service alongside the existing Pursuit for now.

paf31:
This is now running at https://pursuit.purescript.org/

I didn't create a separate URL for this, but it's very simple to switch back to the previous version if necessary.

hdgarrood:
Sorry I wasn't able to find the time this weekend. Looks like it's working well now, though.

PR #232, opened 2016-07-30
Remove search (for now)

paf31:
This is the nuclear option, until we can figure out what's going wrong. This should hopefully reduce the CPU usage for now.

I'll also put an alternative PR together to add logging statements.


hdgarrood:
:+1: lgtm

Issue #226, opened 2016-07-15
High memory usage


paf31:
`top` shows that Pursuit is using ~95% of the CPU on the Linode box, pretty much constantly. I'm not sure if something changed which would cause this, or if it's an issue in some dependency.

hdgarrood:
:( oops. I'm not sure either.

jplatte:
If I `source config/development.env`, `stack exec pursuit` locally it idles at 0% CPU (clicking around a bit in the web browser only slightly increases the mem usage but otherwise does nothing). I only have a database of 4 packages though. Can I somehow download the `data/verified` or all of `data` from pursuit.purescript.org?

hdgarrood:
@jplatte yes, it's all here: https://github.com/purescript/pursuit-backups

hdgarrood:
I have a suspicion that this might be due to the forked Hoogle database regeneration thread getting stuck somehow; this would also explain #221.

paf31:
@hdgarrood Do you think you'll have any time soon to have a look at this with me?

hdgarrood:
Yep. Do you want to set a specific time?

I think I'd start off by aborting a Hoogle database regeneration if it takes more than, say, 30 seconds, and seeing if that makes any difference. I can't think of anything else that would cause this, except for perhaps a bug in Yesod or Warp or something.

paf31:
Does tomorrow work for you? I'll be online tomorrow working on various things, if you have time.

hdgarrood:
Ok, sure, I'll just join IRC whenever I'm available.

paf31:
I looked into this a little. The Hoogle work directory is very full, and after deleting the temp files, I saw some error files with content like this:

```
Parse error 5:2: Unknown attribute: version
Parse error 51:2: Unknown attribute: version
Parse error 158:2: Unknown attribute: version
Parse error 210:9: Parse error: Data.Array.range
Parse error 249:10: Parse error: Data.Array.cons
Parse error 315:10: Parse error: Data.Array.index
Parse error 534:9: Parse error: Data.Array.difference
Parse error 695:2: Unknown attribute: version
Parse error 700:16: Parse error: !
Parse error 742:2: Unknown attribute: version
Parse error 760:10: Parse error: Control.Category.id
Parse error 772:10: Parse error: Control.Biapply.biapply
Parse error 781:10: Parse error: Control.Biapply.biapplyFirst
Parse error 790:10: Parse error: Control.Biapply.biapplySecond
[...]

PR #233, opened 2016-07-31
Add back name search using a trie

### LINKED ISSUE #221: Almost all searches return no results at the moment
At the moment, I'm not getting any results -- as in "No results for ..., sorry." -- for any search, except for searches for package names.

That is, searching for "purescript-lists" works, but searching for "Data.List" does not, and searching for "zip" does not.

paf31:
@hdgarrood Could you please review this? It adds back name search, but we'll need to do more work to implement type search. This should solve a pain point for users for now, at least.

I still need to figure out how to run the license generator.

Fixes #221, #226.


src/Handler/Search.hs
```diff
@@ -0,0 +1,67 @@
...
+
+    jsonOutput = fmap toJSON . traverse searchResultToJSON
+
+searchResultToJSON :: SearchResult -> Handler Value
+searchResultToJSON result@SearchResult{..} = do
+  url <- getFragmentRender <*> pure (routeResult result)
+  return $
+    object [ "package" .= hrPkgName
+           , "version" .= showVersion hrPkgVersion
+           , "details" .= hrDetails
```
paf31: Note this format changed slightly.

hdgarrood:
Looks good 👍

paf31:
Thanks! Obviously, the Hoogle code is still in the Git history, so we can revert back to that when Hoogle 5 is ready, if it's a good fit.

paf31:
Oh, I just realized, I need to update the license file. The executable doesn't work with Stack though. Is there any chance you could please run the script, @hdgarrood?

PR #242, opened 2016-08-13
Add list of latest uploads to home page

### LINKED ISSUE #152: Show most recently uploaded package versions on front page

paf31:
Fixes #152 

@hdgarrood How does this look?

screen shot 2016-08-13 at 12 30 10 pm

### DIFF
--- a/src/Handler/Database.hs
+++ b/src/Handler/Database.hs
@@ -3,6 +3,7 @@ module Handler.Database
   ( createDatabase
   , getAllPackageNames
   , getAllPackages
+  , getLatestPackages
   , lookupPackage
   , availableVersionsFor
   , getLatestVersionFor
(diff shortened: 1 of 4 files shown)


src/Handler/Database.hs
```diff
@@ -33,6 +34,19 @@ getAllPackageNames = do
   contents <- liftIO $ getDirectoryContents (dir ++ "/verified/")
   return . sort . rights $ map mkPackageName contents
 
+getLatestPackages :: Handler [(PackageName, Version)]
+getLatestPackages = do
+    pkgNames <- getAllPackageNames
+    pkgNamesAndVersions <- traverse withTimestamp pkgNames
```
hdgarrood: I might be missing something but could this identifier's name be a little misleading?

templates/default-layout.lucius
```diff
@@ -60,7 +60,7 @@ body {
 
 .col ul,
 .col ol {
-  padding-left: 25px;
+  padding-left: 45px;
```
hdgarrood: Do you know what else is affected by this? I think this affects display of type class members and data constructors, and perhaps also module listings on package homepages, for example. Do they still look ok? Would it make sense to use a class to make this rule a little more specific?

src/Handler/Database.hs
paf31: Oops, yes. I'll fix this.

templates/default-layout.lucius
paf31: I checked, and type classes etc. looked okay to me. I'm happy to revert this though.

hdgarrood: Ok, :+1: to merge then.

PR #245, opened 2016-09-21
fix compilation issues on nixos-16.09

abuibrahim:
With these changes it builds with ghc-7.10.3.

### DIFF
--- a/src/Handler/Database.hs
+++ b/src/Handler/Database.hs
@@ -24,6 +24,8 @@ import Web.Bower.PackageMeta (PackageName, bowerName, bowerDescription,
                               mkPackageName, runPackageName)
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Docs as D
+import Language.PureScript.Docs.RenderedCode.Render (renderType)
+import Language.PureScript.Docs.RenderedCode.Types (RenderedCodeElement(..), outputWith)
 
 import Handler.Utils
 import Handler.Caching (clearCache)
@@ -60,6 +62,15 @@ getAllPackages = do
 tryStripPrefix :: String -> String -> String
 tryStripPrefix pre s = fromMaybe s (stripPrefix pre s)
 
+renderString :: RenderedCodeElement -> String
+renderString str = case str of
+  Syntax s -> s
+  Ident s _ -> s
+  Ctor s _ -> s
+  Kind s -> s
+  Keyword s -> s
+  Space -> " "
+
 createDatabase :: Handler (Trie.Trie [SearchResult])
 createDatabase = do
   pkgs <- getAllPackages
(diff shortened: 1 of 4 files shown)


src/Handler/Database.hs
```diff
@@ -27,6 +27,9 @@ import qualified Language.PureScript.Docs as D
 
 import Handler.Utils
 import Handler.Caching (clearCache)
+import Data.MinLen
+
+type One = Succ Zero
```
paf31: Curious: what was the issue here?

paf31:
Thanks! (although we should figure out why Travis is unhappy obviously 😄)

src/Handler/Database.hs
abuibrahim: Building pursuit-0.4.2...
Preprocessing library pursuit-0.4.2...
[ 6 of 21] Compiling Import.NoFoundation ( src/Import/NoFoundation.hs, dist/build/Import/NoFoundation.o )
Succ and Zero are not re-exported from classy-prelude-1.0.0.1, gettinh this error:
```
Building pursuit-0.4.2...
Preprocessing library pursuit-0.4.2...
[ 6 of 21] Compiling Import.NoFoundation ( src/Import/NoFoundation.hs, dist/build/Import/NoFoundation.o )

src/Import/NoFoundation.hs:11:12:
    Not in scope: type constructor or class ‘Succ’

src/Import/NoFoundation.hs:11:17:
    Not in scope: type constructor or class ‘Zero’
```

After adding `import Data.MinLen` to `src/Import/NoFoundation.hs` getting this:
```
src/Import/NoFoundation.hs:12:1: Warning:
    Defined but not used: type constructor or class ‘One’
``

**Comment by paf31** [comment]
We could try moving to GHC 8 and the corresponding LTS release.

**Comment by abuibrahim** [comment]
Tried bulding with stack resolver set to lts-7.0.

Getting this:
```
cabal: Could not resolve dependencies:
trying: pursuit-0.4.2 (user goal)
trying: base-4.9.0.0/installed-4.9... (dependency of pursuit-0.4.2)
next goal: barrier (dependency of pursuit-0.4.2)
rejecting: barrier-0.1.0 (conflict: base==4.9.0.0/installed-4.9..., barrier =>
base>=4.6 && <4.9)
Dependency tree exhaustively searched.
```

base upper bound needs to be relaxed for barrier I guess

paf31:
Since this is failing still, I think we should try to keep this minimal and stick with GHC 7.10.3 for now. That way, we can hopefully just use `cabal freeze` (or the Stack equivalent, or maybe even just a `cabal.config` based on the LTS version we're using) and run `cabal2nix` and be done. What do you think?

abuibrahim:
I agree. Since not everyone has upgraded to GHC 8, we shouldn't break the build with GHC 7.10.3. Let me see if I can rework the patches.

paf31:
@abuibrahim @hdgarrood I've been thinking about this again. Maybe we should start by using prebuilt binaries in the Nix derivation for Pursuit (and also Try PureScript). I appreciate it's not really in the spirit of Nix, but it would be closer to what we do now, and it would also let us continue using our Stack workflow. And give us faster builds on Nix to boot.

What do you think?

hdgarrood:
My impression was that prebuilt binaries don't work fantastically well on Nix because of shared library paths. If it's possible, though, I'm definitely in favour.

abuibrahim:
There's nothing in the spirit of Nix against pre-built binaries. The problem, as @hdgarrood correctly noted, is shared libs. When you build Pursuit outside of Nix (e.g using stack on a Ubuntu) the binary will be linked against Ubuntu libraries and not those on the Nix machine you want to deploy it to. Technically, you can't even guarantee that Pursuit binary built on one version of Nix will work when simply copied to another.

TryPurescript AFAIK, is mostly a JavaScript/PureScript/JSON bundle with a small scotty based webserver serving it. The javascript part of it can be pre-built and packaged with Nix.

paf31:
Do we use any shared libraries though?

abuibrahim:
GLIBC and whatever else the Haskell RTS links against

hdgarrood:
It may be possible to ask stack to statically link those I think.

PR #246, opened 2016-09-21
Add type to search result

nwolverson:
Adding it to both the HTML output and particularly the JSON response.

As I understand it the JSON result used to contain the `identifier :: Type` as part of the `text` property (which right now seems to be a duplicate of `markup`), which was parsed out in the psc-ide-server https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Ide/Types.hs#L228 - I figure an explicit field with the type is better. Probably this should also show the kind for types etc. And a better field name

### DIFF
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -70,8 +70,8 @@ data SearchResultInfo
   = PackageResult
   | ModuleResult String
   -- ^ Module name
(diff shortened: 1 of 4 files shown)


paf31:
@hdgarrood If this looks okay to you, I can deploy it.

hdgarrood:
LGTM

PR #248, opened 2016-09-27
changes to support building with GHC 8 and the latest LTS

abuibrahim:
These changes fix compilation issues with both GHC 7.10.3 and 8.0.1 on NixOS.
Broken Stack build (classy-prelude incompatible changes) is also fixed with conditional compilation.

### DIFF
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -142,9 +142,9 @@ instance Yesod App where
 
     -- Store session data on the client in encrypted cookies,
     -- default session idle timeout is 120 minutes
-    makeSessionBackend _ = Just <$> defaultClientSessionBackend
+    makeSessionBackend _ = Just <$> envClientSessionBackend
         120    -- timeout in minutes
-        "config/client_session_key.aes"
+        "PURSUIT_CLIENT_SESSION_KEY"
 
     defaultLayout widget = do
         -- We break up the default layout into two components:
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -10,7 +10,7 @@ import Text.Hamlet                 (hamletFile)
 import Text.Jasmine                (minifym)
 import Text.Julius                 (rawJS)
 import Yesod.Core.Types            (Logger)
-import Yesod.Default.Util          (addStaticContentExternal)
+import Yesod.EmbeddedStatic        (EmbeddedStatic, embedStaticContent)
 import qualified Yesod.Core.Unsafe as Unsafe
 import Crypto.Random
 
@@ -109,7 +109,7 @@ instance ToJSON SearchResultInfo where
 -- access to the data present here.
 data App = App
     { appSettings       :: AppSettings
-    , appStatic         :: Static
+    , appStatic         :: EmbeddedStatic
     -- ^ Settings for static file serving.
     , appHttpManager    :: Manager
     , appLogger         :: Logger
@@ -177,24 +177,7 @@ instance Yesod App where
     -- Default to Authorized for now.
     isAuthorized _ _ = return Authorized
 
-    -- This function creates static content files in the static folder
-    -- and names them based on a hash of their content. This allows
-    -- expiration dates to be set far in the future without worry of
-    -- users receiving stale content.
-    addStaticContent ext mime content = do
-        master <- getYesod
-        let staticDir = appStaticDir $ appSettings master
-        addStaticContentExternal
-            minifym
-            genFileName
-            staticDir
-            (StaticR . flip StaticRoute [])
-            ext
-            mime
-            content
-      where
-        -- Generate a unique filename based on the content itself
-        genFileName lbs = "autogen-" ++ base64md5 lbs
+    addStaticContent = embedStaticContent appStatic StaticR minifym
 
     -- What messages should be logged. The following includes all messages when
     -- in development, and warnings and errors in production.
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 module Foundation where
 
 import Import.NoFoundation
--- a/src/Handler/Search.hs
+++ b/src/Handler/Search.hs
@@ -74,7 +74,7 @@ routeResult SearchResult{..} =
       ( PackageVersionModuleDocsR ppkgName pversion modName
       , Nothing
       )
-    DeclarationResult typeOrValue modName declTitle typeText ->
+    DeclarationResult typeOrValue modName declTitle _ ->
       ( PackageVersionModuleDocsR ppkgName pversion modName
       , Just $ pack $ drop 1 $ makeFragment typeOrValue declTitle
       )
--- a/src/Settings.hs
+++ b/src/Settings.hs
@@ -14,6 +14,7 @@ import Network.Wai.Handler.Warp (HostPreference)
 import Yesod.Default.Util (WidgetFileSettings, widgetFileNoReload,
                           widgetFileReload)
 import System.Exit (exitFailure)
+import Yesod.EmbeddedStatic (mkEmbeddedStatic, embedDir)
 
 newtype GithubAuthToken =
   GithubAuthToken { runGithubAuthToken :: ByteString }
@@ -23,9 +24,7 @@ newtype GithubAuthToken =
 -- loaded from various sources: defaults, environment variables, config files,
 -- theoretically even a database.
 data AppSettings = AppSettings
-    { appStaticDir              :: String
-    -- ^ Directory from which to serve static files.
-    , appRoot                   :: Text
+    { appRoot                   :: Text
     -- ^ Base for all generated URLs.
     , appHost                   :: HostPreference
     -- ^ Host/interface the server should bind to.
@@ -62,12 +61,15 @@ data AppSettings = AppSettings
     -- to be uploaded.
     }
 
-isDevelopment :: Bool
 #if DEVELOPMENT
-isDevelopment = True
+#define DEV_BOOL True
 #else
-isDevelopment = False
+#define DEV_BOOL False
 #endif
+mkEmbeddedStatic DEV_BOOL "eStatic" [embedDir "static"]
+
+isDevelopment :: Bool
+isDevelopment = DEV_BOOL
 
 getAppSettings :: IO AppSettings
 getAppSettings = do
@@ -77,7 +79,6 @@ getAppSettings = do
   let appMutableStatic          = isDevelopment
   let appSkipCombining          = isDevelopment
 
-  appStaticDir    <- env "STATIC_DIR" .!= "./static"
   appRoot         <- env "APPROOT" .!= "http://localhost:3000"
   appHost         <- fromString <$> env "HOST" .!= "*4"
   appPort         <- env "PORT" .!= 3000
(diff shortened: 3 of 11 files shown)


src/Foundation.hs
```diff
@@ -172,24 +173,7 @@ instance Yesod App where
...
-            genFileName
-            staticDir
-            (StaticR . flip StaticRoute [])
-            ext
-            mime
-            content
-      where
-        -- Generate a unique filename based on the content itself
-        genFileName lbs = "autogen-" ++ base64md5 lbs
+    addStaticContent = embedStaticContent appStatic StaticR minifym
```
abuibrahim: There are two problems here.

1. Static files are currently not installed. To make sure they are installed by `cabal install` we need to add all static files to the `data-files` section in `pursuit.cabal`
2. The second issue is `addStaticContentExternal` today is going to minify the static assets and drop `tmp/autogen-*` files into the static files directory. This breaks on systems where the installation directory is read-only, like the `/nix/store` on NixOS.

Embedding the static assets into the compiled binary fixes both issues.

src/Foundation.hs
```diff
@@ -137,9 +138,9 @@ instance Yesod App where
 
     -- Store session data on the client in encrypted cookies,
     -- default session idle timeout is 120 minutes
-    makeSessionBackend _ = Just <$> defaultClientSessionBackend
+    makeSessionBackend _ = Just <$> envClientSessionBackend
         120    -- timeout in minutes
-        "config/client_session_key.aes"
+        "PURSUIT_CLIENT_SESSION"
```
abuibrahim: `defaultClientSessionBackend` creates `config/client_session_key.aes` in the installation directory on the first run. This doesn't play well where the installation directory is immutable.
`envClientSessionBackend` allows passing the client session key as an environment variable.

While on subject, I think `PURSUIT_CLIENT_SESSION`  should be changed to `PURSUIT_CLIENT_SESSION_KEY`

abuibrahim:
@hdgarrood, can you please review the changes?

hdgarrood:
Now that I'm back at university I will struggle to find time to do purescript stuff, so for pursuit things I'll generally defer to @paf31 (hope that's ok). Also, stuff that only affects ops (e.g. having an env var for the client session key) I don't mind about, I'll leave that up to @paf31 as he does all the ops stuff.

There's just one thing I would like to say: I don't want to add any more flags to the cabal file or add CPP to support conditional compilation. It should work with what's in the stack.yaml (and updating to LTS 6.18 seems like a good idea) but I really would rather we avoid going to this much effort to support other install plans simultaneously. Really I think we should support only the precise GHC version and install plan specified by stack.yaml.

abuibrahim:
Thanks for reviewing this Harry! So other than facb490 does everything look ok?
@paf31, facb490 is small enough that it can be maintained as a patch to build with Nix.

paf31:
@abuibrahim Presumably we wouldn't need the flag if we do this with a patch file, right? In that case, I think that sounds fine.

I should have some time this week to talk about Nix stuff again, if you like. Sorry, I haven't forgotten, I've just been pretty busy.

src/Settings.hs
```diff
@@ -62,6 +61,13 @@ data AppSettings = AppSettings
     -- to be uploaded.
     }
 
+#if DEVELOPMENT
+#define DEV_BOOL True
+#else
+#define DEV_BOOL False
+#endif
+mkEmbeddedStatic DEV_BOOL "eStatic" [embedDir "static"]
```
hdgarrood: I'd rather keep the amount of CPP as low as possible. This looks fine but if we are adding it, then I think I'd rather modify the definition of `isDevelopment` to use it too.

src/Handler/Search.hs
```diff
@@ -59,7 +59,7 @@ routeResult SearchResult{..} =
       ( PackageVersionModuleDocsR ppkgName pversion modName
       , Nothing
       )
-    DeclarationResult typeOrValue modName declTitle typeText ->
+    DeclarationResult typeOrValue modName declTitle _typeText ->
```
hdgarrood: Is this change because it was generating an unused binding warning? If that's the case I think I'd rather just use `_`, since I don't think the leading underscore approach is very well known (in fact, I had to go and look it up just now).

src/Import/NoFoundation.hs
```diff
@@ -1,12 +1,9 @@
...
+    ( module Import
     ) where
 
 import ClassyPrelude.Yesod   as Import
 import Settings              as Import
 import Yesod.Core.Types      as Import (loggerSet)
 import Yesod.Default.Config2 as Import
 import Control.Category      as Import ((>>>), (<<<))
-
-type One = Succ Zero
```
abuibrahim: This change fixes the following compilation error on NixOS:

```
Building pursuit-0.4.7...
Preprocessing library pursuit-0.4.7...
[ 6 of 21] Compiling Import.NoFoundation ( src/Import/NoFoundation.hs, dist/build/Import/NoFoundation.o )

src/Import/NoFoundation.hs:12:12: error:
    Not in scope: type constructor or class ‘Succ’

src/Import/NoFoundation.hs:12:17: error:
    Not in scope: type constructor or class ‘Zero’

```

src/Handler/Search.hs
abuibrahim: I wasn't sure if _typeText was intended to be used in the future so left it as a (sort of) comment.  Can certainly change it to _.

src/Settings.hs
abuibrahim: Done.

src/Import/NoFoundation.hs
hdgarrood: Is this because you're building with a different set of dependencies than what is specified in stack.yaml?

abuibrahim: Yes, NixOS Haskell dependencies are bit different from stackage LTS. This change makes it possible to build with both.

src/Handler/Search.hs
hdgarrood: If it turns out we want to use it in the future then we can always change it back to `typeText` without the leading underscore (and in that case I'd rather we did do that than using the leading underscore version).

abuibrahim: Makes sense. I have pushed the change to _ as you suggested.

src/Import/NoFoundation.hs
hdgarrood: Right, ok. I would prefer to keep exporting `One` from here, I think. If there's no neat way of doing that then I think I'd rather not include this change in this pull request.

It has just also occurred to me that some form of the `maximum` function should be exported from here too, so I'd rather not have the explicit MinLen import in whichever other module it was.

abuibrahim: All right then. I am going to take this change out, so the rest can be merged.

hdgarrood:
Are we considering using NixOS for the community server, by the way?

paf31:
> Are we considering using NixOS for the community server, by the way?

Yes, @abuibrahim has done some work on Nix-ifying Pursuit, and I quite like the idea of having the config under a GitHub repo where I can take PRs to deploy user's services.

paf31:
Plus the ability to run locally or deploy to AWS, maintain mirrors easily, etc.

hdgarrood:
Right, yeah, that does all sound like it would be really useful. Regarding the changes regarding dependencies which I asked to be removed, then, is there not some way to package a Haskell package for NixOS that just uses the dependencies specified by its stack.yaml? &mdash; I think it's quite important to have the same install plans in development and production.

paf31:
I'm not sure if there is such a thing as `stack2nix` yet, like there is `cabal2nix`. I think one option is to develop using Cabal inside a Nix shell, but that's not ideal if people are used to Stack.

hdgarrood:
Come to think of it, we don't want to be compiling Pursuit on the community server at all, do we? Can we just keep on compiling it on Travis like we do now and publishing binaries on the Releases tab on github and create a Nix package that just downloads it from there?

paf31:
We certainly don't want to build on the community server, but hopefully we could have Nix build in a VM and push the closure to the target machine, or something like that.

abuibrahim:
@paf31 that's right. And you can then use NixOps to deploy it to EC2, GCE or Azure.

abuibrahim:
BTW, stack supports Nix (I haven't tried it though).
https://github.com/commercialhaskell/stack/blob/master/doc/nix_integration.md

hdgarrood:
Is building on a different OS going to be awkward because of shared library paths?

abuibrahim:
Nix builds are completely isolated from host OS. It uses http://nixos.org/patchelf.html to modify the dynamic linker and RPATH of the final executable. Works like a charm.

hdgarrood:
Sorry, I mean compiling on an OS other than Nix and then running on Nix; is that awkward and if so, why?

abuibrahim:
Oh, I see what you mean. No, your executables built on Ubuntu will not work on NixOS, unless statically linked.

hdgarrood:
Right, ok. And is there a way to tell Nix that we depend on a set of libraries from Hackage at specific versions, instead of just specifying that we depend on the libraries and getting whatever versions of them are in `nixpkgs` (or wherever)? `stack list-dependencies` prints this information in a fairly easy to consume format:

```
$ stack list-dependencies
Glob 0.7.5
HUnit 1.3.1.1
ReadArgs 1.2.2
StateVar 1.1.0.4
adjunctions 4.3
aeson 0.11.2.0
aeson-better-errors 0.9.0.1
aeson-compat 0.3.3.0
aeson-pretty 0.7.2
ansi-terminal 0.6.2.3
ansi-wl-pprint 0.6.7.3
appar 0.1.4
array 0.5.1.0
[...]
```

abuibrahim:
Nix used to follow stackage LTS releases, but it has been discontinued. You can see the announcement and and reasons here http://lists.science.uu.nl/pipermail/nix-dev/2016-June/020585.html

hdgarrood:
Surely there's a way to use a specific set of libraries just for when you're inside a particular project, though?

abuibrahim:
Surely there is :). You start by doing cabal2nix to generate your nix file. You can choose different versions of packages availble in nixpkgs. 

Here's are the packaged versions of yesod-forms for GHC8.0.1

```
$ nix-env -qaP -A nixos.haskell.packages.ghc801 yesod-form
nixos.haskell.packages.ghc801.yesod-form        yesod-form-1.4.7.1
nixos.haskell.packages.ghc801.yesod-form_1_4_8  yesod-form-1.4.8
```
and for GHC7.10.3
```
$ nix-env -qaP -A nixos.haskell.packages.ghc7103.yesod-form
nixos.haskell.packages.ghc7103.yesod-form  yesod-form-1.4.7.1
```
You can also override the packages with your own versions if needed.

 The Haskell development workflow is very well documented here
 http://nixos.org/nixpkgs/manual/#users-guide-to-the-haskell-infrastructure

I propose we take the discussion of Nix Haskell workflow off of this thread, since all the Nix specific changes have been reverted, as per your request. The current PR will compile and work fine with Stack and developers can continue to use stack as before. The small NixOS specific changes can be maintained as patches for those who would like to deploy it with NixOps.

hdgarrood:
@paf31 I would quite like to merge this now, actually, as this fixes some compiler warnings, and I'd rather avoid conflicts. Is that ok? I think the only necessary change to the deployment setup is to put the client session key into an environment variable instead of the file `config/client_session_key.aes` (but it would also be good to remove the stuff that copies static files around since that's no longer needed).

paf31:
@hdgarrood Sounds good, I'll merge and try to get this deployed.

PR #250, opened 2016-11-15
Type search

paf31:
This uses a heuristic to search types, based on a notion of closeness of two types.


hdgarrood:
Seems sensible enough to me :+1:

PR #256, opened 2016-12-08
Add explanations for why readmes are missing

### LINKED ISSUE #161: Display why readmes were not available
I am aware of two potential causes: The tag doesn't exist on github, or Pursuit was rate limited. We've seen both of these before and it's helpful to know which one (even though I doubt we'll see rate limiting again).

hdgarrood:
Fixes #161


paf31:
👍 Great!

PR #257, opened 2016-12-09
Use latest classy Prelude via extra-deps

paf31:
This lets us use the same dependencies as we're using for the Nix build, which means we don't need to maintain a patch file for that deployment.

The `stack.yaml` file is a bit ugly right now, but I think we can update to LTS 7.12 as soon as `barrier` gets updated for GHC 8 (there is an open issue [here](https://github.com/philopon/barrier/issues/3))

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -129,6 +129,8 @@ library
                  , xss-sanitize
                  , wai-middleware-gunzip
                  , barrier ==0.1.*
+                 , minlen
+                 , mono-traversable
 
     if flag(dev)
       build-depends: foreign-store
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -16,7 +16,6 @@ import Crypto.Random
 
 import Web.Bower.PackageMeta (PackageName, parsePackageName, runPackageName)
 import qualified Data.Trie as Trie
-import Control.DeepSeq (NFData, rnf)
 import Data.Version
 import Model.DocLinks (TypeOrValue(..))
 import qualified Css
--- a/src/Handler/Database.hs
+++ b/src/Handler/Database.hs
@@ -14,6 +14,7 @@ module Handler.Database
 import Import
 import qualified Data.Aeson as A
 import qualified Data.ByteString.Lazy as BL
+import qualified Data.MinLen as MinLen
 import qualified Data.Text as T
 import qualified Data.Trie as Trie
 import Data.Version (Version, showVersion)
@@ -29,6 +30,8 @@ import Language.PureScript.Docs.RenderedCode.Types (RenderedCodeElement(..), out
 
 import Handler.Utils
 import Handler.Caching (clearCache)
+
+type One = MinLen.Succ MinLen.Zero
 
 getAllPackageNames :: Handler [PackageName]
 getAllPackageNames = do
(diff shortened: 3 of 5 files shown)


hdgarrood:
I think we can drop the MinLen dependency by using the `maximum` from [Data.NonNull](https://hackage.haskell.org/package/mono-traversable-1.0.1/docs/Data-NonNull.html) instead (and I think this `maximum` is re-exported by ClassyPrelude too).

Also, dealing with dependencies this way makes me worry that they'll get out of sync. Once nixpkgs is updated, we will still be using the dependencies listed here. I'd rather do it the other way around, i.e. use the dependencies specified in our stack.yaml when building for Nix, and ignore the default versions of Haskell libraries that Nix provides.

paf31:
Actually, `nix-pkgs` is ahead of Stackage in the sense that it already contains an updated `barrier` (via a patch file). Once Stackage gets `barrier`, we can update to LTS 7.12, remove these `extra-deps`, and the two builds should be identical.

hdgarrood:
Sure, but we still have no reason to expect Stackage LTS 7 to stay in sync with what's in nixpkgs, right? I still think it makes the most sense to stick with Stack/Stackage and tell our Nix script to use what's in stack.yaml (i.e. don't get Haskell library dependencies from nixpkgs), so that there's one definitive location for pursuit's dependency information.

paf31:
> tell our Nix script to use what's in stack.yaml

How would we do that?

hdgarrood:
I don't know, I am not a Nix user. I would be very surprised if it's not possible, though.

paf31:
> we still have no reason to expect Stackage LTS 7 to stay in sync with what's in nixpkgs, right?

I think nix-pkgs in version 16.09 is supposed to track that version of LTS. @abuibrahim Can you confirm?

abuibrahim:
@paf31: Nix "loosely" follows the LTS and by loosely they mean

> where "loosely" means that we'll honor the mandated version bounds for
libraries but tend to ignore them for executables.

abuibrahim:
@hdgarrood: The dependency information comes from the .cabal file not from stack.yaml. cabal2nix can then be used to turn it into Nix "derivation". The build can still fail if the library bounds do not match what's in nix store. If that happens typically a Nix user would open a ticket with upstream package and temporarily add a patch.

paf31:
@hdgarrood From the small amount of experience I have with Nix, I really think that if we want to use Nix at all, then this is the best option for now. Syncing up Stack dependencies with Nix on the Nix side is going to be a maintenance nightmare.

If you like, I can show you my local deployment of Pursuit inside VirtualBox using @abuibrahim's Nix script, so you can see it working.

hdgarrood:
If Nix makes it so difficult to do an entirely reasonable thing such as this, I worry that there might be other entirely reasonable things we want to do that would become very difficult if we were using Nix.

Anyway, didn't @abuibrahim suggest that it would be possible [in this comment](https://github.com/purescript/pursuit/pull/248#issuecomment-263761434)? I did a little searching and found [a guide to using Stack to build Haskell projects](http://nixos.org/nixpkgs/manual/#how-to-build-a-haskell-project-using-stack), too.

paf31:
I haven't tried using stack's support for Nix, I'll try that soon.

paf31:
Let's handle this on the Nix side as a separate thing. We can use Stack here, and we'll deal with the incompatibilities using patch files until Pursuit can use LTS 7.12 and GHC 8.

paf31:
@hdgarrood What about forking `barrier`, updating to LTS 7.12 and using the fork for now?

hdgarrood:
Yep, that sounds good to me.

paf31:
Ok, I'll work on that later then, thanks.

PR #258, opened 2016-12-10
LTS 7.12

paf31:
Uses my branch of `barrier` for now, until the upstream PR gets merged.

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -129,6 +129,8 @@ library
                  , xss-sanitize
                  , wai-middleware-gunzip
                  , barrier ==0.1.*
+                 , minlen
+                 , mono-traversable
 
     if flag(dev)
       build-depends: foreign-store
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -129,7 +129,6 @@ library
                  , xss-sanitize
                  , wai-middleware-gunzip
                  , barrier ==0.1.*
-                 , minlen
                  , mono-traversable
 
     if flag(dev)
--- a/src/Application.hs
+++ b/src/Application.hs
@@ -1,3 +1,6 @@
+-- To work around a bug in GHC 8.0.1:
+-- See https://groups.google.com/forum/#!topic/yesodweb/DlyXqFM7ZnY
+{-# LANGUAGE NoDisambiguateRecordFields #-}
 {-# LANGUAGE PackageImports #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 module Application
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -1,3 +1,6 @@
+-- To work around a bug in GHC 8.0.1:
+-- See https://groups.google.com/forum/#!topic/yesodweb/DlyXqFM7ZnY
+{-# LANGUAGE NoDisambiguateRecordFields #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 module Foundation where
 
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -16,7 +16,6 @@ import Crypto.Random
 
 import Web.Bower.PackageMeta (PackageName, parsePackageName, runPackageName)
 import qualified Data.Trie as Trie
-import Control.DeepSeq (NFData, rnf)
 import Data.Version
 import Model.DocLinks (TypeOrValue(..))
 import qualified Css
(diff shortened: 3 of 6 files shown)


paf31:
This is ready for review again.

hdgarrood:
Hm, I'm not overly keen on switching to GHC 8.0.1 if it's going to require a workaround like this (and especially not a workaround for a bug that makes GHC panic). Also to be clear on what I said before, I have no problem with switching to LTS 7 in principle but I would still rather avoid Nix until we find a way to get Nix to use the exact same dependencies as what is specified by a stack.yaml file (if that's what has prompted this).

paf31:
You mean `NoDisambiguateRecordFields`? Well, that's fixed in 8.0.2, which is due out already, so it should be fixed very soon.

I very much doubt we'll find a nice way to get Nix to use `stack.yaml` any time soon, but I'd like a way to deploy Pursuit using Nix rather than running a bunch of one-off scripts, so it'd be good to figure this out if we can.

hdgarrood:
Ok, then let's go ahead with this PR and Nix; I've made my case, so since I haven't been able to convince you I'll drop it. :+1:

paf31:
Well I really don't want the Nix deployment work to get in the way of regular development, but I don't think it should. It's possible that we could transition to working in a Nix shell or something eventually, but for now I think it's reasonable for us to use Stack for development and for me to use Nix separately for deployment. It's just that there are going to be these overlapping concerns, like choosing approximately equal dependencies. Personally, if we have to pick, I'd rather use a slightly buggy GHC8 and up-to-date libraries.

hdgarrood:
Oh sure, absolutely - I don't expect this to get in the way of regular development at all, it's just the risk that stuff will behave slightly differently in production that I was thinking about.

naehrstoff:
Just FYI: I tried to get this repo running on macOS 10.12 today and failed miserably (GHC panic), even when using this branch. What helped in the end was the following `stack.yml` that my colleague put together for me based on a [discussion in Stack](https://github.com/commercialhaskell/stack/issues/2577#issuecomment-264403014).

```
resolver: lts-7.12
compiler-check: match-exact
compiler: ghc-8.0.1.20161117
setup-info:
  ghc:
    linux64:
      8.0.1.20161117:
        url: http://downloads.haskell.org/~ghc/8.0.2-rc1/ghc-8.0.1.20161117-x86_64-deb8-linux.tar.xz
        content-length: 112047972
        sha1: 6a6e4c9c53c71cc84b6966a9f61948542fd2f15a
    macosx:
      8.0.1.20161117:
        url: https://downloads.haskell.org/~ghc/8.0.2-rc1/ghc-8.0.1.20161117-x86_64-apple-darwin.tar.xz
        content-length: 113379688
        sha1: 53ed03d986a49ea680c291540ce44ce469514d7c
    windows64:
      8.0.1.20161117:
        url: https://downloads.haskell.org/~ghc/8.0.2-rc1/ghc-8.0.1.20161117-x86_64-unknown-mingw32.tar.xz
        content-length: 155652048
        sha1: 74118dd8fd8b5e4c69b25df1644273fbe13177c7
flags:
  pursuit:
    dev: true
packages:
- '.'
- location:
    git: https://github.com/philopon/barrier.git
    commit: 22d9041bd01f4590f517ad7b426c665104eefd97
extra-deps:
- wai-middleware-gunzip-0.0.2
```

paf31:
Thanks @naehrstoff!

@hdgarrood What do you think about using the release candidate instead?

hdgarrood:
Sounds good to me :+1:

hdgarrood:
If we switch to the GHC 8.0.2 release candidate can we remove the LANGUAGE pragmas for working around that bug?

paf31:
I believe so, yes.

hdgarrood:
Great. This would also be useful for what I'm working on at the moment, so could I merge this now and make that change?

hdgarrood:
BTW It looks like barrier 0.1.1 has been released too with your changes in, so we could just change that to an extra-dep now.

paf31:
@hdgarrood Sounds good to me, thanks.

hdgarrood:
I've reverted back to GHC 8.0.1, as Travis couldn't deal with using the GHC 8.0.2 RC. If anyone else feels like investigating this please feel free, but I am not going to do anything about it because I suspect it will 'just work' once GHC 8.0.2 proper is released.

Issue #109, opened 2015-07-11
Favicon


hdgarrood:
See also purescript/purescript.github.io#21

danse:
:+1: 

I think that this is even more important here, because the Pursuit page is a search engine. I use to keep it opened and pinned together with other pages useful when developing, and the icon is the only way one can distinguish them

texastoland:
If you tell me what to do with it I can PR.

danse:
:raised_hands:

Issue #93, opened 2015-07-02
Insufficient contrast (according to WCAG)


hdgarrood:
See: http://webaim.org/resources/contrastchecker/

The following have insufficient contrast, according to WCAG:
- Links: background #FFFFFF, foreground #C4943A
- Keywords in code like "forall", "instance": background #F2F5F9, foreground #A0A0A0

Presumably the links issue also affects the main PureScript website as well.

Unfortunately if you simply darken the link colour until it passes, it becomes a rather horrible brown. So I'm not really sure how to proceed here.

naehrstoff:
I experimented with finding a good color for the links using this [contrast finder tool](http://contrast-finder.tanaguru.com/result.html?foreground=%23C4943A&background=%23FFFFFF&isBackgroundTested=false&ratio=4.5&algo=Rgb).

However, what happens when we increase the contrast of the links, is that they become less distinguishable from the rest of the text, i.e. we fix readability but hurt discoverability 😏

I think having the links in bold at 16px (with a darker color on hover in https://github.com/purescript/pursuit/pull/266) is sufficient for the time being. I solved the issue with the keyword color in https://github.com/purescript/pursuit/pull/266/commits/40ee7cdbb1337b9177b04322833982e8f1abbffd.

A color scheme redesign should take this accessibility issue into account. For the time being, I think this is good enough while allowing us to keep a yellow tone (the most difficult color to get good contrast on!) and I suggest to close the issue.

PR #266, opened 2017-01-03
UI Refinements

naehrstoff:
This PR focuses on making Pursuit more useable and beginner friendly by applying a few visual and interaction design recipes. It mainly improves on the following:

* More welcoming homepage with more personality, search examples, and denser presentation of packages.
* Distinctive typographic treatment of the content to provide better visual guidance for the reader.
* Improved display of search results for easier scannability and understandability (e.g. by adding badges to modules and packages to better highlight what they are).
* Clarified navigation between packages and modules and allow navigation from one module to a sibling module.
* Friendlier display of code examples and inline code snippets.

I'll have some more time to work on this this week, so if there are things that I should fix or that I overlooked, please let me know and I'll see that I can add them so this PR could be merged.

_Note: this is a big PR that I should probably have announced before I started working on it. I chose to do this in stealth mode, however, because I was not sure whether I would find the time to properly work on it due to the unpredictability of kids 😬. It did work out and I'm happy to be able to submit this PR._

# Commit summary

## Typography
* Fixed display of italic fonts (no italic font files had been loaded before, making the browser slant the regular font, leading to faux-italics). faux-real-italic
* Lists align with left edge for cleaner look.
* Introduced definition lists for combination of titles and list elements.
* Version numbers are displayed in a smaller font to make them more distinctive.

## CSS

I refactored the CSS quite a bit fully aware that large scale code rewriting is a very bad practice for PRs. However, after working on the previous CSS for a while the many global styles made it very difficult to work with, leading me to believe that it was grown over time and was ready for a refactoring.

I introduced code sections and a naming system based on [BEM](https://en.bem.info/). I understand that this is a controversial naming system; it is very good at creating human-readable namespaces, however, which is needed for creating CSS components that don't influence one another and that can be changed confidently over time. Because it only has three rules (that I applied here) it is easy to apply also for people who don't know what it is.

I tried to keep the CSS small and well structured to make sure other people can make changes to it easily.

* Introduced code sections
* Component naming system based on BEM
* Generalised utility classes
* Extracted existing styles into CSS components
* Improved page layout at various breakpoints
* Treat Github rendered Markdown content separately and clearly namespace it (because we don't control it)

## Other

* Fixed some small accessibility issues

# Discussion

1. [ ] I saw that there is an open issue for a favicon (#109), work on it has been started but stalled. If it's ok, I would make and add a favicon.
2. [ ] Issue #93 mentions contrast of links and other elements. This is something that could be attacked in further design refinements. However, it opens the question on what decisions taken there would mean for the Purescript website. Would that maybe be an undertaking for the Purescript 1.0 release? I like to use living style guides to make sure a design language can grow and is properly applied, but I wouldn't know where to host that (A separate repo? The main Purescript repo?)
3. [ ] A recent PR (https://github.com/purescript/pursuit/commit/27e0ac19026538b04929837c7f9bf5fb1ebe2cb7) introduced a new route to display builtin docs. I have for the time being solved it like so in the new layout: https://github.com/naehrstoff/pursuit/commit/cc9ee90efc8c8fd2c4b36f92cc466f5472f4a6ae Is more needed?
4. [x] After merging in the latest changes I found that some packages did not render or stopped rendering after a while (e.g. purescript-ace). I'm not sure what the cause of this is but suspect that it has to do with the `purescript-backups` packages I use (updated just now) or some caches. If you can reproduce this, please let me know.

# Screenshots

## Homepage

homepage

## Package

package

## Module

module

## Search results

search-results

## Code

code

### DIFF
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -212,13 +212,12 @@ documentationPage ::
 documentationPage pkg@D.Package{..} widget =
   let pkgName = D.packageName pkg
   in [whamlet|
-    <div .clearfix>
-      <div .col.col--main>
-        <h1>
-          package
-          <a href=@{packageRoute pkg}>#{runPackageName pkgName}
-
-      ^{versionSelector pkgName pkgVersion}
+    <div .page-title .clearfix>
+      <h1 .page-title__title>
+        <span .package-label>Package
+        <a href=@{packageRoute pkg}>#{runPackageName pkgName}
+      <div .page-title__version-selector>
+        ^{versionSelector pkgName pkgVersion}
 
     <div .col.col--main>
       ^{widget}
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -163,10 +163,12 @@ getBuiltinDocsR mnString = do
         setTitle (toHtml mnString)
         let mn = P.moduleNameFromString mnString
         let htmlDocs = primDocs
-        let widget = $(widgetFile "packageVersionModuleDocs")
         [whamlet|
-          <div .col.col-main>
-            ^{widget}
+          <div .col.col--main>
+            <div .page-title.clearfix>
+              <div .page-title__label>Builtin
+              <h1 .page-title__title>#{insertBreaks mn}
+            #{htmlDocs}
           |]
     _ ->
       defaultLayout404 $ [whamlet|
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -145,6 +145,7 @@ getPackageVersionDocsR (PathPackageName pkgName) (PathVersion version) =
 getPackageVersionModuleDocsR :: PathPackageName -> PathVersion -> Text -> Handler Html
 getPackageVersionModuleDocsR (PathPackageName pkgName) (PathVersion version) mnString =
   cacheHtml $ findPackage pkgName version $ \pkg@D.Package{..} -> do
+    moduleList <- renderModuleList pkg
     mhtmlDocs <- renderHtmlDocs pkg mnString
     case mhtmlDocs of
       Nothing -> notFound
@@ -152,8 +153,7 @@ getPackageVersionModuleDocsR (PathPackageName pkgName) (PathVersion version) mnS
         defaultLayout $ do
           let mn = P.moduleNameFromString mnString
           setTitle (toHtml (mnString <> " - " <> runPackageName pkgName))
-          documentationPage pkg $
-            $(widgetFile "packageVersionModuleDocs")
+          $(widgetFile "packageVersionModuleDocs")
 
 getBuiltinDocsR :: Text -> Handler Html
 getBuiltinDocsR mnString = do
@@ -206,22 +206,6 @@ versionSelector pkgName version = do
   let route = PackageAvailableVersionsR (PathPackageName pkgName)
   availableVersionsUrl <- getUrlRender <*> pure route
   $(widgetFile "versionSelector")
-
-documentationPage ::
-  D.VerifiedPackage -> WidgetT App IO () -> WidgetT App IO ()
-documentationPage pkg@D.Package{..} widget =
-  let pkgName = D.packageName pkg
-  in [whamlet|
-    <div .page-title .clearfix>
-      <div .page-title__label>Package
-      <h1 .page-title__title>
-        <a href=@{packageRoute pkg}>#{runPackageName pkgName}
-      <div .page-title__version-selector>
-        ^{versionSelector pkgName pkgVersion}
-
-    <div .col.col--main>
-      ^{widget}
-    |]
 
 uploadPackageForm :: Html -> MForm Handler (FormResult FileInfo, Widget)
 uploadPackageForm = renderDivs $ areq fileField settings Nothing
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -213,14 +213,14 @@ documentationPage pkg@D.Package{..} widget =
   let pkgName = D.packageName pkg
   in [whamlet|
     <div .clearfix>
-      <div .col.col-main>
+      <div .col.col--main>
         <h1>
           package
           <a href=@{packageRoute pkg}>#{runPackageName pkgName}
 
       ^{versionSelector pkgName pkgVersion}
 
-    <div .col.col-main>
+    <div .col.col--main>
       ^{widget}
     |]
 
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -166,7 +166,7 @@ getBuiltinDocsR mnString = do
         [whamlet|
           <div .col.col--main>
             <div .page-title.clearfix>
-              <div .page-title__label>Builtin
+              <div .page-title__label>Module
               <h1 .page-title__title>#{insertBreaks mn}
             #{htmlDocs}
           |]
--- a/src/Handler/Packages.hs
+++ b/src/Handler/Packages.hs
@@ -213,8 +213,8 @@ documentationPage pkg@D.Package{..} widget =
   let pkgName = D.packageName pkg
   in [whamlet|
     <div .page-title .clearfix>
+      <div .page-title__label>Package
       <h1 .page-title__title>
-        <span .package-label>Package
         <a href=@{packageRoute pkg}>#{runPackageName pkgName}
       <div .page-title__version-selector>
         ^{versionSelector pkgName pkgVersion}
--- a/templates/default-layout.lucius
+++ b/templates/default-layout.lucius
@@ -150,11 +150,6 @@ a:hover {
   text-decoration: underline;
 }
 
-a.weight-normal,
-a.weight-normal:hover {
-  font-weight: normal;
-}
-
 h1 {
   font-size: 3.814em;
   font-weight: 300;
@@ -410,9 +405,6 @@ p {
  * Font sizes are based on a major third scale (1:1.250)
  * ========================================================================== */
 
-a h2 { /* FIXME: target with class name instead */
-  margin-top: 0.83em;
-}
 
 ul {
   list-style-type: none;
--- a/templates/default-layout.lucius
+++ b/templates/default-layout.lucius
@@ -462,7 +462,7 @@ ol li {
 
 .decl__body .keyword,
 .decl__body .syntax {
-  color: #a0a0a0;
+  color: #0B71B4;
 }
 
 
--- a/templates/default-layout.lucius
+++ b/templates/default-layout.lucius
@@ -2,9 +2,7 @@
  * ========================================================================== */
 
 html {
-  overflow-y: scroll;
   box-sizing: border-box;
-  color: #000;
 }
 
 *, *:before, *:after {
@@ -13,6 +11,7 @@ html {
 
 body {
   background-color: #{Css.backgroundColor};
+  color: #000;
   font-family: "Roboto", sans-serif;
   line-height: 1.563em;
 }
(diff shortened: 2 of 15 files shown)


hdgarrood:
Wow, this looks really nice, thanks! I'll try to review in the next few days.

hdgarrood:
> After merging in the latest changes I found that some packages did not render or stopped rendering after a while (e.g. purescript-ace)

Were you seeing JSON parsing errors in the logs? If so, that's probably https://github.com/purescript/purescript/issues/2530. Hopefully there will soon be a 0.10.5 release which restores backwards compatibility. In the meantime you could manually regenerate docs for any packages that you want to use while developing - I wrote [a script](https://github.com/hdgarrood/dotfiles/blob/4ae0f56ae1dcdfac9e0f58e6084841474df2d6f8/bin/upload-pursuit) to make this easier because I do this fairly often.

naehrstoff:
That's the error, yes. I just tried to get your script – or at least the `psc-publish` part – running, but failed because I don't have all the tools installed globally (I have per-project setups using Boxen). But considering it works for some packages (e.g. purescript-thermite) but doesn't for others (e.g. purescript-ace) I think it's safe to assume that the 0.10.5 release you mention will fix this.

paf31:
This looks great!

hdgarrood:
Regarding the favicon issue (thanks for reminding me), we do have one, we just haven't gotten around to modifying the HTML to actually use it on purescript.org yet. I've just opened a new PR on the purescript.org repo for that: https://github.com/purescript/purescript.github.io/pull/105. If you would like to address the pursuit favicon issue in this PR &mdash; which would be lovely &mdash; you could use the file from that PR.

templates/default-layout.lucius
```diff
@@ -153,226 +159,671 @@ html, body {
...
 }
 
 td, th {
-  padding: 0.4em;
+  text-align: left;
+  padding: 0.41em 0.51em;
 }
 
-tbody tr:nth-child(odd) {
-  background-color: #eee;
```
hdgarrood: Is this rule preserved in the current CSS? I'd quite like to keep it as I think it's very helpful for reading tables. For example: https://pursuit.purescript.org/packages/purescript-sequences/1.0.2

templates/homepage.hamlet
```diff
@@ -1,30 +1,53 @@
...
-    packages, see the <a href=@{HelpR}>Help page</a>.
+    Pursuit hosts API documentation for <a href="http://www.purescript.org/">PureScript</a> packages. It lets you search by package, module, and function names, as well as approximate type signatures.
+
   <p>
-    Pursuit is free and open-source software, and the code is hosted on GitHub.
-    Contributions are welcome: <a href="https://github.com/purescript/pursuit">https://github.com/purescript/pursuit</a>.
+    To get started, have a look around the <a href="/packages/purescript-prelude">Prelude</a> – PureScript’s standard library – or try one of these example searches:
+
+  <ul .ul--search>
+    <li><a href="/search?q=map">map</a></li>
```
hdgarrood: This is a great thing to have on the homepage :+1: :+1: :+1:

src/Handler/Packages.hs
```diff
@@ -163,10 +163,12 @@ getBuiltinDocsR mnString = do
         setTitle (toHtml mnString)
         let mn = P.moduleNameFromString mnString
         let htmlDocs = primDocs
-        let widget = $(widgetFile "packageVersionModuleDocs")
         [whamlet|
-          <div .col.col-main>
-            ^{widget}
+          <div .col.col--main>
+            <div .page-title.clearfix>
+              <div .page-title__label>Builtin
```
hdgarrood: I think this should say "Module" rather than "Builtin" - `Prim` is still a module, it being built in to the compiler doesn't change that.

templates/default-layout.lucius
```diff
@@ -1,30 +1,76 @@
...
+ **
+ ** Components can be nested.
+ ** ************************************************************************* */
+
+
+/* Section: Document Styles
+ * ========================================================================== */
+
 html {
-  overflow-y: scroll;
```
hdgarrood: I'd quite like to keep this too, as without it everything will shift slightly to the side when moving from a page which isn't large enough to generate a scrollbar to one that is.

hdgarrood:
Actually don't worry about rebasing, I've done it for you, and pushed the result to the `ui-refinements` branch in this repository.

src/Handler/Packages.hs
hdgarrood: Other than that the builtin docs route looks great! :+1:

templates/homepage.hamlet
```diff
@@ -1,30 +1,53 @@
...
-        <li>
-          <a href=@{PackageVersionR (PathPackageName pkgName) (PathVersion version)}>
-            #{runPackageName pkgName} - #{showVersion version}
-    <h2>Packages:
-      $forall pkgs <- pkgNamesByLetter
-        $forall letter <- headMay pkgs >>= firstLetter
-          <h3>#{letter}
+    <div .multi-col .clearfix>
+      <div .multi-col__col>
+        <h3>Latest packages
```
hdgarrood: I think the previous text "Latest uploads" would be more appropriate here.

hdgarrood:
I did some testing on mobile and found a slight issue with the display of the version selector:

screenshot from 2017-01-05 00-49-40

hdgarrood:
The above image is me testing in Responsive Design mode in desktop firefox, but I observed the same issue in iOS safari

hdgarrood:
On second thoughts, let's leave the favicon out of this PR and address it separately. This is very close to being mergeable in my mind so I'd like to merge it soon :)

I think the only things you raised which I haven't yet responded to are the large scope of the CSS refactoring and the living style guide / design language question. With respect to the CSS refactoring, I'm 100% in favour, this is clearly a big improvement. Regarding the other question, I'm not totally sure what the phrases 'living style guide' or 'design language' mean, although I can make an educated guess. I think that sort of thing would be ideally placed in a separate repository. I don't think it's absolutely necessary to have one, but if you wanted to work on that it would certainly be appreciated! Again I think we can address that outside of this PR.

hdgarrood:
To summarise, here's the complete list of I think needs doing before we can merge this:

* [ ] Fix the warning and make CI pass
* [ ] Change "Latest Packages" to "Latest Uploads"
* [ ] Fix overlapping text issue on mobile, perhaps by moving the repository text down a little on narrower devices?
* [ ] Decide whether to keep the `html { overflow-y: scroll; }` rule
* [ ] Decide whether to keep the `tbody tr:nth-child(odd) { background-color: #eee; }` rule

hdgarrood:
And a list of things we might want to address later:

* Favicon
* Living style guide, probably in a separate repo

naehrstoff:
Thanks for this thorough review, @hdgarrood! I'll answer to some comments now and will make the changes by tomorrow morning so we have this PR in a mergeable state.

templates/default-layout.lucius
naehrstoff: I removed it in order to use the design pattern with the horizontal lines a bit more, but I have no problem adding this again. It'll look like in the attached screenshot.

Even/Odd:
with

Current:
without

src/Handler/Packages.hs
naehrstoff: Sure, I didn't quite understand this :)

templates/homepage.hamlet
hdgarrood: Ahh, right, that makes sense. FWIW, the reason I prefer "latest packages" is that in my mind, the word "package" doesn't really refer to a moment in time, whereas "upload" does, and so referring to them as "latest" doesn't quite work. Come to think of it, I think "most recently uploaded packages" would be OK, and it's hopefully then clear that all the other things are actually packages?

templates/default-layout.lucius
hdgarrood: Hm, actually that doesn't work so well since there's no padding on the left hand side really, does it? Come to think of it the current one is actually perfectly readable as is, IMO. Shall we leave this out after all?

templates/homepage.hamlet
hdgarrood: Also to answer your other question, yes, I think ideally "uploader" is the person who uploaded it, and "publisher" is the person who created the git tag (although I don't think it's all that easy for us to work out who created the git tag right now).

templates/default-layout.lucius
naehrstoff: Ok, then I would suggest to keep it as is now so that we have a calmer left spine of the text body (the padding would break that). It's an easy fix if you discover you would still like to have the stripes.

templates/homepage.hamlet
naehrstoff: Yeah, I copied the idea from Hoogle :)

templates/default-layout.lucius
hdgarrood: :+1:

naehrstoff:
Alright, I fixed all the issues you found. I wasn't able to check the todos on your list, so here they are again:

* [x] Fix the warning and make CI pass
* [x] Change "Latest Packages" to "Latest Uploads" **→ Added a title for the overall section to strike a balance with labeling things and making headings not too long so they would wrap around. Please also have a look at the commit messages in #74ec201b and #d37a70c1**
* [x] Change the "Builtin" text on the builtin docs page to "Module" **→ Renamed**
* [x] Fix overlapping text issue on mobile, perhaps by moving the repository text down a little on narrower devices?  **→ Fixed through more robust layout for small screens**
* [x] Decide whether to keep the html { overflow-y: scroll; } rule **→ Added rule and comment**
* [x] Decide whether to keep the tbody tr:nth-child(odd) { background-color: #eee; } rule **→ No even/odd style**

Regarding the living style guide: this is a tool I would recommend for a project like Purescript where several different code bases (the project website, TryPurescript, Pursuit, …) should speak the same design language (i.e. look and feel similar to communicate to the user that they belong together). Such a style guide is used to collect all the components used in all the websites to make sure that lists, tables, buttons, etc. all look the same across projects. It's also used for collecting decisions on colours, typography, behaviours, etc. We use this tool a lot in our work and found it very useful (we're using our own [Catalog](https://interactivethings.github.io/catalog/)). An example of such a style guide is Google's [Material Design](https://material.io/guidelines/), and there's many more resources on http://styleguides.io

Anyways, my holidays are over and I won't be able to work on a style guide for a while 😅 But if you want to keep this in mind for a Purescript 1.0 release I'd be happy to talk again about this/help out later this year or so.

Let me know if there's anything else you want me to do for this PR 😄

hdgarrood:
Thanks again! This looks really fantastic.

templates/default-layout.lucius
```diff
@@ -23,6 +25,16 @@
...
+ **
+ **
+ ** TYPOGRAPHY
+ **
+ ** Typographic choices for sizes, line-heights and margins are based on a
+ ** musical major third scale (4:5). This gives us a way to find numbers
+ ** and relationships between them that are perceived as harmonic.
+ **
+ ** To make use of this modular scale, follow this formula:
+ **   (5/4)^n | n ∈ [-6, 8]
```
hdgarrood: I'm just wondering, does this mean that n is an integer between -6 and 8?

naehrstoff: I guess it's difficult to understand without an example? Maybe we could also add this link that shows the relationship:

http://www.modularscale.com/?1&em&1.25&js&text

Or write out the actual numbers: `…, 0.64, 0.8, 1, 1.25, 1.563, …`

Should I try to add something like this? (And where, now that it's been merged (yay!)).

hdgarrood: I just wanted to check about the notation, since n ∈ [a, b] usually means that n is a real number such that a <= n and n <= b. Up to you, you could send another PR if you wanted to. :)

PR #269, opened 2017-01-06
Fix inconsistent paddings in code signatures

noraesae:
The original paddings for `result__signature` and `decl__signature` have different top and bottom padding values. This can be seen in the following screenshots.

#### result__signature

screen shot 2017-01-07 at 03 06 16

#### decl__signature

screen shot 2017-01-07 at 02 49 26

## After

After fixing them, they look equal on both top and bottom.

#### result__signature

screen shot 2017-01-07 at 03 42 34

#### decl__signature

screen shot 2017-01-07 at 03 44 10

### DIFF
--- a/templates/default-layout-wrapper.hamlet
+++ b/templates/default-layout-wrapper.hamlet
@@ -6,7 +6,7 @@ $doctype 5
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>#{pageTitle'}
 
-    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,400i,700,700i" type="text/css" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css?family=Roboto+Mono|Roboto:300,400,400i,700,700i" type="text/css" rel="stylesheet">
     <link href=@{StaticR css_normalize_css} type="text/css" rel="stylesheet">
     <script type="text/javascript" src=@{StaticR js_js_cookie_js}>
     <script type="text/javascript" src=@{StaticR js_Markup_js}>
--- a/templates/default-layout.lucius
+++ b/templates/default-layout.lucius
@@ -202,7 +202,7 @@ code, pre {
   background-color: #{Css.codeBackground};
   border-radius: 3px;
   color: #{Css.codeForeground};
-  font-family: Consolas, "Liberation Mono", Menlo, Courier, monospace;
+  font-family: "Roboto Mono", monospace;
   font-size: 87.5%;
 }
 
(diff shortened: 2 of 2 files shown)


hdgarrood:
Thanks, but I'm not sure that the paddings necessarily need to be equal on the top and the bottom; I think it's probably okay as it is. Maybe @naehrstoff would like to comment?

noraesae:
Yes, it is not a big problem, but I guess it could look awkward to some people(including me).

If @naehrstoff intentionally made the paddings different, I am not strongly against it. Please consider accepting this PR when everybody thinks it's okay.

naehrstoff:
Thanks for caring, @noraesae! Your screenshots look ok, but when I try these values on my Mac with Chrome/Firefox/Safari it doesn't look good anymore:

pursuit-padding

It seems that your browser is using a different font. Compare the letter shapes that I circled: this is not the same font.

pursuit-font-comparison

This makes sense, because the font stack is specified like so: `Consolas, "Liberation Mono", Menlo, Courier, monospace` (as used by Github, for example), so it's allowed to pick a different font if you don't have Consolas installed. And different fonts have different metrics, which is why we see this weird padding behaviour. What OS/browser are you using?

This leaves us with these options:

1. Leave the padding as it was and only cater to users who have Consolas installed.
2. Use a different font stack.
3. Use a monospaced font from Google Fonts (like we do with Roboto) to make sure everyone gets the same experience.

Option 1 is not very friendly towards all users, option 2 does not look very good (not many good default monospaced fonts that are available on all OSes), option 3 increases the page weight by some kB.

I'm currently in favor of option 3 and would suggest to use [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono). This would be included like this: `<link href="https://fonts.googleapis.com/css?family=Roboto+Mono|Roboto:300,400,400i,700,700i" rel="stylesheet">` and then used like this: `font-family: 'Roboto Mono', monospace;`. And it would look like this:

roboto-mono

What do you think? Would it be possible for you to add these changes? If we add Roboto Mono, your settings for the padding are correct.

noraesae:
As @naehrstoff  guessed, I do not have Consolas installed, so it worked with a fallback, Menlo. I have tested with latest Chrome, Safari and Firefox in macOS 10.12.

I also think 3 is the best way to handle this. I have just uploaded an additional commit to add Roboto Mono. It does not really increase page size too much (as suggested, some KB) and now it is ensured every user has the same experience. The font itself is good as well. When offline, it will use a system monospace fallback.

#### result__signature

screen shot 2017-01-08 at 12 24 08

#### decl__signature

screen shot 2017-01-08 at 12 25 01

#### fallback

screen shot 2017-01-08 at 12 25 30

hdgarrood:
Looks great, thanks both!

PR #272, opened 2017-01-07
Add favicon

### LINKED ISSUE #109: Favicon

naehrstoff:
This commit uses a 32x32 PNG favicon instead of an ICO because the ICO turned out to be much bigger (15kB for the ICO vs 350B for the PNG, of course the ICO includes sizes 16/32/48) due to the alpha transparency we want to use.

Alternatively, we could just serve multiple PNGs like we do on the Purescript website. Here I failed in figuring out how to do this with Yesod :(

Preferrably, we would also include touch icons and the like to cater for all platforms. However, here too I do not know how to do that with Yesod (we'd have to have routes to SVG, JSON, XML, PNG, and ICO).

Closes #109


hdgarrood:
Unfortunately this isn't the clearest thing ever but one of the Template Haskell invocations, either `mkYesodData` or `mkYesodDispatch` (can't remember which), defines a value for each file in the static directory. So for example we have a file called `static/css/normalize.css`, and this turns into a value called `css_normalize_css`. Then you can use the `StaticR` constructor to turn one of these values into a proper route. See, for example, in `default-layout-wrapper.hamlet` we have this:

    <link href=@{StaticR css_normalize_css} type="text/css" rel="stylesheet">

This is, I think, what we want to do to support multiple favicons.

hdgarrood:
So to clarify, I think all we need to do is:

* Put each of the favicon files into a new `static/img` directory
* Add `<link rel=icon>` elements in `default-layout-wrapper.hamlet` using `StaticR` routes.

It would also be nice to remove that `FaviconR` route while we're at it :)

naehrstoff:
I tried routes like `img_favicon_ico` but that didn't work. But I would like to do this the right way instead of just the way that works now, so I'd like to fix this. It will likely be next weekend, as a pretty stressful week is coming up, but I'll keep this on the radar!

hdgarrood:
Did you get an error message? Perhaps I can help diagnose the problem.

kRITZCREEK:
would be nice if we could make this happen :)

naehrstoff:
Yes, I'm sorry it hasn't happened yet – family and business got in the way (and will do for a while still). It's one of those seemingly small things that should take 5 minutes – but then there are other such small things that fill up all the space …

Anyway, it's not forgotten and the files are ready to be integrated, I hope I can do it soon.

hdgarrood:
Also, where did that safari-pinned-tab SVG come from?

naehrstoff:
I want to make another two commits tonight, one will be documentation on how to make the favicon assets (but I think I'll add that to the purescript-logo repo).

I'm currently at ZuriHac.info, so I finally got some time for this again, yay :)

naehrstoff:
Please review, @hdgarrood. I added more icons for more use-cases in separate commits because you might prefer not to include all the changes, as they require adding new routes to Yesod. However, I would recommend to add the full set because Pursuit is a reference documentation and when people save the URL to their home screens now they will get a nice logo on all devices and browsers.

Please let me know if I should change anything in the Yesod code so I can fix it and we can finally close this PR after a long, long while 😄

P.S. The build doesn't run because there's a dependency on the `alex` binary. I don't know what it does, but I installed it with `stack install alex` and it worked from there on. This might be something to add to the readme.

hdgarrood:
Great! Let's come back to this after https://github.com/purescript/logo/pull/6 is sorted and merged, maybe?

hdgarrood:
Ok actually ignore my last comment, I can review some of this now.

* I would rather not have documentation in the `static` directory, as it doesn't really make sense for us to be hosting that markdown file on pursuit, and also I don't want to duplicate this documentation between here and the logo repo. I'd rather just have a note somewhere in this repo pointing to the logo repo, and telling people to go there for documentation.
* I'm not really keen on putting files like `manifest.json` and `browserconfig.xml` into Haskell code like you've done here. I suppose there is some type safety benefit, but I think it's outweighed by the fact that we have duplicated information. I'd rather just store it in the usual format of an xml or json file, i.e. how it's stored in the logo repo (or, how it will be stored after your other PR is merged).
*  Is it necessary to have separate routes for browserconfig.xml; can they not just go under `static` like everything else/

naehrstoff:
Thanks for all the feedback, will get to it!

Regarding the `manifest.json` and `browserconfig.xml`: I could not figure out another way to use `StaticR` in these files to point to the built files. I wanted to use `StaticR` even though it seems that the resulting paths (`static/res/…`) are predictable – but they might differ during development or could change if the routing changes somehow and we wouldn't really notice … (But I don't know enough Yesod to be sure of the behaviour here.)

I'll gladly rip out this code – I learned a lot of Haskell researching how to do this, but I agree that it is information that shouldn't really go in code :) What would be the approach you would take to solve this?

Regarding the route for `browserconfig.xml`: two reasons. One is that I didn't know how else to add a handler for it :) and two is that some Internet Explorers apparently look for this file in the root. But as we are linking to it from the `<head>`, it should be possible to remove this, I can research this again.

hdgarrood:
> I wanted to use `StaticR` even though it seems that the resulting paths (`static/res/…`) are predictable – but they might differ during development or could change if the routing changes somehow and we wouldn't really notice … (But I don't know enough Yesod to be sure of the behaviour here.)

I am fairly certain that the only way this would change is if we consciously changed it, so I don't think this is a problem. I suppose it's possible (although very unlikely) that updating to a new version of Yesod would change the routes, but in any case the compiler will check that the URLs we list in the `head` element won't 404.

> I'll gladly rip out this code – I learned a lot of Haskell researching how to do this, but I agree that it is information that shouldn't really go in code :) What would be the approach you would take to solve this?

I think it would be best to wait until your logo PR is merged and then just copy all the necessary files into a subdirectory of `static/`.

> Regarding the route for `browserconfig.xml`: two reasons. One is that I didn't know how else to add a handler for it :) ...

Ah right, yeah. We shouldn't need a handler for it if we have it as a static file instead of having it in Haskell code.

> and two is that some Internet Explorers apparently look for this file in the root. But as we are linking to it from the `<head>`, it should be possible to remove this, I can research this again.

Ok great, let me know what you find out :)

naehrstoff:
Alright, I ripped out the whole routing machinery. Still happy I was able to get it working, but also happy to not have so much more code :)

Regarding making `browserconfig.xml` available from the root of the website: some IEs will check for the existence of that file anyway, so they'll get a 404. This is a wasted request, but it's safe to ignore. Something similar happens with `favicon.ico`: if it's linked to from the `<head>` some browsers might choose to use it instead of `favicon_x_.png`, which will lead to blurry images. But the way it's setup now should be fine for most browsers – and the worst that can happen is that it'll be slightly blurry, which is fine.

So, from my point of view this is fine as it is now, unless you see something more to change :)

hdgarrood:
LGTM, thanks very much!

paf31:
@hdgarrood Just let me know if/when you'd like me to make a release.

hdgarrood:
Ok! I thought I'd wait a bit as I was hoping to fix a couple more bugs over the next day or two.

PR #283, opened 2017-02-13
Update to purescript 0.10.7

hdgarrood:
Also update to LTS 8. The latest ClassyPrelude stopped exporting `IOData` which necessitated making a few minor changes / switching lazy bytestrings for strict ones (which is probably what we should be doing anyway, as these json blobs are all going to be pretty small).

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -100,7 +100,7 @@ library
                  , containers
                  , vector
                  , time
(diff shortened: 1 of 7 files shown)


hdgarrood:
@paf31 I'm going to interpret your thumbs-up as a review :smile:

Issue #132, opened 2015-07-19
Searching by type is too general


Thimoteus:
Suppose I want to find a function with type signature Char -> String, but I don't know if it's called toString or fromChar or something else. If I search for `Char -> String` in pursuit, it doesn't show -- in fact, the second result is unsafeCoerce.

hdgarrood:
Still appears to be an issue after #235. `Data.String.singleton` is the 10th result for the query `Char -> String`; arguably it should be the first.

hdgarrood:
Fixed by @matthewleon in #286 - thanks!

PR #285, opened 2017-02-18
search: promote exact type matches

matthewleon:
addresses #132

(Data.String.singleton should be the first result for the query "Char -> String")

This is probably not the ideal solution to the issue, but in lieu of an overhauled type matching algorithm, it seems to work as a pragmatic stopgap.

### DIFF
--- a/src/Handler/Search.hs
+++ b/src/Handler/Search.hs
@@ -28,7 +28,7 @@ getSearchR = do
     Nothing -> redirect HomeR
     Just query -> do
       results <- case tryParseType query of
-        Just ty | not (isSimpleType ty) -> searchForType ty
+        Just ty | not (isSimpleType ty) -> searchForType ty query
         _ -> searchForName (toLower query)
       selectRep $ do
         provideRep (htmlOutput query results)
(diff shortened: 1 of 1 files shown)


hdgarrood:
The type matching algorithm isn't particularly complicated right now, so I'd prefer to do it properly - which means first finding out exactly why `Char -> String` doesn't give `singleton` the highest score. Have you been able to work out why that is?

matthewleon:
It does give singleton the highest score, which is zero, for a complete match. It just gives a whole lot of other things a zero as well, and then sorts alphabetically.

I can go have another look to establish a more type-based way to "punish" everything that isn't an _exact_ type match by raising its score. That is, indeed, the proper way to do it.

That said, even if I were to do that, is there any actual harm in raising the score of everything that doesn't have a string type match? It seems to me that, at the very least, it can act as a "safety net" to ensure that even if there is a flaw in our logic that we don't figure out, the match the user is looking for gets a fighting chance of making it to the top.

A hack, admittedly, but I think it's a very pragmatic one.

hdgarrood:
To be able to convincingly argue that it's pragmatic, we'd first need to establish that doing it properly is prohibitively expensive; that hasn't happened yet as far as I can see.

matthewleon:
I'll give it another shot.

hdgarrood:
Thanks :smile:

PR #286, opened 2017-02-18
Improve compareTypes algorithm.

matthewleon:
addresses #132

Comparisons between `TypeVar`s and something that is _not_ a type variable should be penalized by at least 1 point. This ensures that a query like `Char -> String` only awards a 0 point result to the correct `singleton` function, or that `Int -> Number` yields `toNumber` first.

(previously, matches including type variables like `forall a b. a -> b` would be rewarded 0 points for both of these queries.)

### DIFF
--- a/src/Handler/Search.hs
+++ b/src/Handler/Search.hs
@@ -105,7 +105,7 @@ searchForType ty = do
     -- or Nothing otherwise. Lower scores are better.
     compareTypes :: P.Type -> P.Type -> Maybe Int
     compareTypes (P.TypeVar _) (P.TypeVar _) = Just 0
-    compareTypes t (P.TypeVar _) = Just (typeComplexity t)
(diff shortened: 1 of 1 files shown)


matthewleon:
Some more handy results:

before PR: `Int -> Number` just yields a list of `a -> b` functions sorted alphabetically
after PR: `Int -> Number` yields `toNumber` first, `timeStamp` (`forall e. IsEvent e => e -> Number`) second, then the aforementioned list.

before PR: `a -> a -> Boolean` yields a list of `a -> b -> c` functions sorted alphabetically
after PR: a list of `a -> b -> Boolean` functions, sorted alphabetically, at the top.

A future PR could potentially ensure that in the last search, `a -> a -> Boolean` matches over `a -> b -> Boolean`. This would be a bit trickier, though.

paf31:
👍 I like it. Thanks @matthewleon!

PR #288, opened 2017-03-09
Use LESS for CSS instead of Lucius

hdgarrood:
Makes it easier to re-use this CSS in other projects by using LESS for
variables instead of Lucius.

For now I've just checked in the generated CSS, which will then be embedded into the `pursuit` executable. At some point in the future we might want to consider a more sophisticated build process but I'd rather not fiddle around with something like that unless I have a reason to.

The reason I would like to do this is that I am looking at generating HTML docs inside `purs docs`, and I would like to reuse this CSS.

/cc @naehrstoff in case you are interested.


paf31:
Looks nice 👍 Should I run the `less` command when I deploy this next then?

hdgarrood:
No need: I didn't want to add any compile time dependency on node.js or anything so I just ran `lessc` locally and checked in the result as `static/css/pursuit.css`.

PR #294, opened 2017-03-31
Updates for 0.11

paf31:
I can deploy this over the weekend if it looks okay. I think we might be able to keep the existing packages which is nice.


paf31:
I'll make a release, and do a test deployment to see if the `verified` files still show up. If they don't, I'll roll back.

PR #296, opened 2017-04-08
Use purescript-0.11.3

paf31:
I thought this would fix the issue I saw when uploading `arrays`:

```
* At the path: ["modules"][0]["declarations"][19]["info"]["type"]
Arising from an Aeson FromJSON instance:
key "contents" not present
```

assuming it was caused by [my recent PR](https://github.com/purescript/purescript/commit/c82defc6f1ded53f42a49c7507561ef1feb6a359).

But now I see that change has been around since 0.11.0, so would have been in the previous Pursuit build. Also, I don't think I touched any code which would affect that path in the JSON, although I might be mistaken.

Any thoughts, @hdgarrood?

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -1,5 +1,5 @@
 name:              pursuit
-version:           0.5.0
+version:           0.6.0
 cabal-version:     >= 1.8
 build-type:        Simple
 license:           MIT
@@ -99,7 +99,7 @@ library
                  , containers
                  , vector
                  , time
-                 , purescript >= 0.11.1
+                 , purescript >= 0.11.3
                  , bower-json >= 1.0.0.1
                  , blaze-builder
                  , blaze-markup
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -86,7 +86,7 @@ library
                  , directory                     >= 1.1        && < 1.3
                  , warp                          >= 3.0        && < 4
                  , data-default
-                 , aeson                         >= 0.11       && < 0.12
+                 , aeson
                  , conduit                       >= 1.0        && < 2.0
                  , monad-logger                  >= 0.3        && < 0.4
                  , fast-logger                   >= 2.2        && < 2.5
--- a/stack.yaml
+++ b/stack.yaml
@@ -14,5 +14,5 @@ extra-deps:
 - optparse-applicative-0.13.2.0
 - pipes-4.2.0
 - pipes-http-1.0.5
-- purescript-0.11.1
+- purescript-0.11.3
 - websockets-0.9.8.2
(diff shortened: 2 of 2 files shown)


paf31:
Actually, it looks like the issue might be here:

https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/Types.hs#L780

Since the parser uses `type` and the encoder uses `data`.

paf31:
Okay, I think I know what's going on here. I dumped the JSON and formatted it with `jq`:

```
$ purs publish --manifest bower.json --resolutions /tmp/resolutions.json | jq '.modules[0].declarations[19].info.type'
Warnings:                                                              
  The following installed packages were missing path information in the
  resolutions file:                                                    
    * purescript-console                                               
    * purescript-assert                                                
    * purescript-const                                                 
{
  "tag": "ForAll",
  "contents": [
    "a",
    {
      "tag": "TypeApp",
      "contents": [
        {
          "tag": "TypeApp",
          "contents": [
            {
              "tag": "TypeConstructor",
              "contents": [
                [
                  "Prim"
                ],
                "Function"
              ]
            },
            {
              "tag": "TypeApp",
              "contents": [
                {
                  "tag": "TypeConstructor",
                  "contents": [
                    [
                      "Prim"
                    ],
                    "Array"
                  ]
                },
                {
                  "tag": "TypeVar",
                  "contents": "a"
                }
              ]
            }
          ]
        },
        {
          "tag": "TypeApp",
          "contents": [
            {
              "tag": "TypeConstructor",
              "contents": [
                [
                  "Data",
                  "Maybe"
                ],
                "Maybe"
              ]
            },
            {
              "tag": "TypeApp",
              "contents": [
                {
                  "tag": "TypeConstructor",
                  "contents": [
                    [
                      "Prim"
                    ],
                    "Record"
                  ]
                },
                {
                  "tag": "RCons",
                  "contents": [
                    "head",
                    {
                      "tag": "TypeVar",
                      "contents": "a"
                    },
                    {
                      "tag": "RCons",
                      "contents": [
                        "tail",
                        {
                          "tag": "TypeApp",
                          "contents": [
                            {
                              "tag": "TypeConstructor",
                              "contents": [
                                [
                                  "Prim"
                                ],
                                "Array"
                              ]
                            },
                            {
                              "tag": "TypeVar",
                              "contents": "a"
                            }
                          ]
                        },
                        {
                          "tag": "REmpty"
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    },
    null
  ]
}
```

Everything here has a `contents` field, with the exception of `REmpty`. I'm guessing this is related to some change in the way `aeson` handles nullary data constructors by default, perhaps.

hdgarrood:
I suppose we should bump the lower bound on aeson in purescript up to 1.0?

paf31:
Yes, I think so. Although it's pinned by LTS for now at least.

I'll merge this and try deploying again, thanks.

PR #302, opened 2017-04-24
Better 'package not found' message

hdgarrood:
An example of the message this might now generate:

>  It might not have been published to Pursuit as of the currently active version series of the PureScript compiler (namely, 0.11.x).

I think this is an improvement as the current message suggests that there might be a bug in Pursuit, which has turned out to almost never be the case; rather, it's almost always because the package hasn't been reuploaded since the last breaking change.


paf31:
👍 Looks good!

PR #309, opened 2017-06-11
Clarify documentation with regards to building

naehrstoff:
It was not clear before how to develop incrementally and add data to the database, now all development related activities are grouped under a "Development" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,16 +51,17 @@ To run the web server on <http://localhost:3000>:
 $ stack exec pursuit
 ```
 
-Make sure you have content in your `data/verified` directory (see
-[Database](#database)), otherwise you will not be able to browse any packages. The
-database will be regenerated from this data source before the server starts
-listening; this can take a short time depending on how much data you have.
+You might want to add some content to the database (see [Database](#database)),
+otherwise you will not be able to browse any packages. The database will be
+regenerated from this data source before the server starts listening; this
+can take a short time depending on how much data you have.
 
 ## Database
 
 Pursuit currently uses the filesystem as a database, since it requires no setup
 and it makes it easy to use Git and GitHub for backing up. The data directory
-is set via an environment variable (see [Configuration](#configuration)).
+is set via an environment variable (see [Configuration](#configuration), the
+default is `data`).
 
 If you need some sample packages to work with, you can clone the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,37 +14,64 @@ Information for package authors can be found at
...
 ```
-$ stack build --flag pursuit:-dev
+$ stack ghci
 ```
 
-## Database structure
+Once the REPL has loaded, you can reload the code and then update the web server:
+
+```
+> :l DevelMain
```
hdgarrood: :+1: I never bothered to work out how to do this, just hit Ctrl-C and then ran `stack build && stack exec pursuit`.

README.md
```diff
@@ -14,37 +14,64 @@ Information for package authors can be found at
...
+
+### Web server
+
+To run the web server on <http://localhost:3000>:
+
+```
+$ stack exec pursuit
+```
+
+Make sure you have content in your `data/verified` directory (see
```
hdgarrood: Can we please remove this reference to "the `data/whatever` directory"? Since `./data` is just the default value of the configuration variable for the data directory this isn't necessarily true.

README.md
```diff
@@ -14,37 +14,64 @@ Information for package authors can be found at
...
+If you need some sample packages to work with, you can clone the
+[pursuit-backups][pursuit-backups] repo and copy the packages you want to the
+`verified/` directory. This is more convenient than manually uploading each
+package.
+
+[pursuit-backups]: https://github.com/purescript/pursuit-backups
+
+## Database structure
+
+The database lives in the `data` directory, its structure is as follows:
```
hdgarrood: I think this needs to be amended slightly as it seems to contradict the sentence in a previous paragraph:

> The data directory is set via an environment variable (see [Configuration](#configuration)).

README.md
hdgarrood: Also I think this is a little too strong. I'd prefer something like "You might want to add some content" rather than "Make sure you have content".

README.md
naehrstoff: That took several minutes for me, so I was glad that @justinwoo had this nice trick in store for me 😄

hdgarrood:
Thanks!

PR #313, opened 2017-06-24
Update deps to include happy & alex

AidanDelaney:
Trivial dependency patch.


hdgarrood:
I don't think this will have the desired effect, unfortunately.

AidanDelaney:
Gah, it won't. Sorry.

PR #316, opened 2017-07-11
Redirect URLs without a version to the latest version

### LINKED ISSUE #277: Rewrite URLs that point to docs with no specific version to redirect to the latest version
This would be super useful for when making links from non-pursuit documentation, as in the Halogen guide for instance.

The idea being, I could link to:

```
https://pursuit.purescript.org/packages/purescript-halogen/docs/Halogen.Query#v:action
```

Which would (currently) redirect to:

```
https://pursuit.purescript.org/packages/purescript-halogen/0.12.0/docs/Halogen.Query#v:action
```

It's not without flaws, as obviously the docs can still end up out of date, referring to dead links, but at least it avoids the overhead of having to update every link in the documentation before a new release, pointing to URLs that don't exist (yet, as updating the docs is part of the release).

hdgarrood:
Fixes #277

I tested this and it does include the fragment too, so e.g. if you use the URL that @garyb gave

> https://pursuit.purescript.org/packages/purescript-halogen/docs/Halogen.Query#v:action

it does focus on `action` after redirecting to the page with the latest version. :tada:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
   - mkdir -p ~/.local/bin
   - export PATH=$HOME/.local/bin:$PATH
   - curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
+  - stack setup
   - stack --version
   - stack install alex happy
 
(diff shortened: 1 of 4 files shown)


garyb:
Awesome, thanks!

fiatjaf:
Is this live?

hdgarrood:
No, if an issue is closed that just means the fix is in master. If you want to know what's live then you should compare the version at the bottom of any page on pursuit.purescript.org with the git history here

Issue #305, opened 2017-05-19
don't limit search results


cdepillabout:
It appears that Pursuit limits search results to a fixed number.

For instance, when I search for ["encode"](https://pursuit.purescript.org/search?q=encode), it doesn't return the result `encodeURIComponent`.  It does, however, return this result when I search for something more specific like ["encodeuri"](https://pursuit.purescript.org/search?q=encodeuri).

It would be nice to use paging to return _all_ results.

Or, at the very least, it would be nice to add a little message to the bottom of the search results page saying that there may be more results that aren't shown.

hdgarrood:
:+1: I'd like to investigate paging results.

PR #317, opened 2017-07-12
WIP: Implement search results pagination

felixSchl:
Stab at #305

It's working but the HTML is pretty horrifying for large result sets as there's a link to every single page in the navigation bar. I was thinking this would be a nice way to render the navigation: https://stackoverflow.com/a/14193775/2954634?

Also, in order to know the total amount of pages, I removed the [`take 50`](https://github.com/purescript/pursuit/blob/c16211e916b52f446d9f7ae4dce967629ffac0d5/src/Handler/Search.hs#L90), so it's doing a lot more work now in case of large result sets. If so desired, we could simply take 51, then check if the result is greater than 50, then simply render a "next" link, as opposed to individual page links.

Would be great to get some feedback on the above so I can adjust the code accordingly

### DIFF
--- a/src/Handler/Search.hs
+++ b/src/Handler/Search.hs
@@ -7,6 +7,10 @@ module Handler.Search
 import Import
 import Data.Trie (elems, submap)
 import Data.Version (showVersion)
+import qualified Data.Map.Strict as Map
+import qualified Data.Text as T
+import Text.Read (readMaybe)
+import Control.Monad.Trans.Maybe (runMaybeT, MaybeT(..))
 import qualified Web.Bower.PackageMeta as Bower
 
 import Language.PureScript.Docs.AsHtml (makeFragment, renderMarkdown)
@@ -21,22 +25,53 @@ import qualified Language.PureScript as P
 
 import qualified XMLArrows
 
+resultsPerPage :: Int
+resultsPerPage = 5
+
 getSearchR :: Handler TypedContent
 getSearchR = do
   mquery <- lookupGetParam "q"
+  page <- maybe 1 (max 1) . ((readMaybe . T.unpack) =<<) <$> lookupGetParam "page"
+
   case mquery of
     Nothing -> redirect HomeR
     Just query -> do
-      results <- case tryParseType query of
+      (results, nPages) <- applyPagination page <$> case tryParseType query of
         Just ty | not (isSimpleType ty) -> searchForType ty
         _ -> searchForName (toLower query)
       selectRep $ do
-        provideRep (htmlOutput query results)
+        provideRep (htmlOutput query results page nPages)
         provideRep (jsonOutput results)
   where
-    htmlOutput :: Text -> [SearchResult] -> Handler Html
-    htmlOutput query results = do
+    applyPagination :: Int -> [a] -> ([a], Int)
+    applyPagination page xs =
+      let zs = drop ((page - 1) * resultsPerPage) xs
+          tot = length xs
+          nPages = tot `div` resultsPerPage
+       in (take resultsPerPage zs, nPages)
+
+    mkPaginationW :: Int -> Int -> Handler (Maybe Widget)
+    mkPaginationW page nPages = runMaybeT $ do
+      cr <- MaybeT getCurrentRoute
+      MaybeT $ do
+        urp <- getUrlRenderParams
+        getParams <- Map.fromList . reqGetParams <$> getRequest
+        let mkPageLink :: Int -> Text
+            mkPageLink p = urp cr $ do
+                Map.toList $ Map.insert "page" (pack $ show p) getParams
+            hasNextPage = page < nPages
+            hasPrevPage = page > 1
+            allPages :: [Int] = [1 .. nPages]
+
+        if nPages <= 1
+            then pure Nothing
+            else pure . Just $ do
+              $(widgetFile "pagination")
+
+    htmlOutput :: Text -> [SearchResult] -> Int -> Int -> Handler Html
+    htmlOutput query results page nPages = do
       fr <- getFragmentRender
+      mPaginationW <- mkPaginationW page nPages
       content <- defaultLayout $(widgetFile "search")
       sendResponseStatus ok200 content
 
@@ -87,12 +122,12 @@ routeResult SearchResult{..} =
 searchForName :: Text -> Handler [SearchResult]
 searchForName query = do
   db <- atomically . readTVar =<< (appDatabase <$> getYesod)
-  return (map fst (take 50 (concat (elems (submap (encodeUtf8 query) db)))))
+  return $ map fst $ concat $ elems $ submap (encodeUtf8 query) db
 
 searchForType :: P.Type -> Handler [SearchResult]
 searchForType ty = do
     db <- atomically . readTVar =<< (appDatabase <$> getYesod)
-    return (map fst (take 50 (sortBy (comparing snd) (mapMaybe (matches ty) (concat (elems db))))))
+    return $ map fst $ sortBy (comparing snd) (mapMaybe (matches ty) $ concat (elems db))
   where
     matches :: P.Type -> (a, Maybe P.Type) -> Maybe (a, Int)
     matches ty1 (a, Just ty2) = do
--- a/static/css/pursuit.css
+++ b/static/css/pursuit.css
@@ -619,6 +619,18 @@ ol li {
 .result__actions__item + .result__actions__item {
   margin-left: 0.65em;
 }
+/* Component: Pagination
+ * -------------------------------------------------------------------------- */
+ul.pagination li {
+  float: left;
+  padding-left: 0.6em;
+}
+ul.pagination li.active {
+  text-decoration: underline;
+}
+ul.pagination li::before {
+  content: inherit;
+}
 /* Component: Version Selector
  * -------------------------------------------------------------------------- */
 .version-selector {
--- a/templates/search.hamlet
+++ b/templates/search.hamlet
@@ -1,6 +1,9 @@
 <div .col.col--main>
   <h1>Search results
 
+  $maybe paginationW <- mPaginationW
+    ^{paginationW}
+
   $if null results
     <div .result.result--empty>
       Your search for <strong>#{query}</strong> did not yield any results.
@@ -44,3 +47,7 @@
           <span .result__actions__item>
             <span .badge.badge--module title="Module">M
             #{moduleName}
+
+  $maybe paginationW <- mPaginationW
+    ^{paginationW}
+
--- a/templates/pagination.hamlet
+++ b/templates/pagination.hamlet
@@ -1,0 +1,21 @@
+<ul .pagination>
+    <li .previous :not hasPrevPage:.disabled>
+        $if hasPrevPage
+            <a href="#{mkPageLink $ page - 1}">
+                &larr;
+        $else
+            <span .disabled>
+              &larr;
+
+    $forall p <- allPages
+        <li :p == page:.active>
+            <a href="#{mkPageLink p}">
+                #{show p}
+
+    <li .next :not hasNextPage:.disabled>
+        $if hasNextPage
+            <a href="#{mkPageLink $ page + 1}">
+                &rarr;
+        $else
+            <span>
+              &rarr;


hdgarrood:
In my mind it's a lot more important that the load on the server is kept down than the user knows how many pages there are; from that perspective I don't think we can justify calculating the total amount of pages immediately. I can't imagine a context in which it would be useful to know how many pages of results there are.

How about the first page displays the first 50 results, as it does currently, but then, if there are more results, there is a 'next page' link which takes you to page 2, and then if there are still more, there is a 'next page' link for page 3, and so on?

I think we should still have a cutoff point where the server refuses to do any more work to find results. I think 10 pages is probably a reasonable cutoff point -  if there are still more results after that I think it's safe to say it's unlikely they're going to be useful, and so instead of rendering a 'next page' link, we could have a message that indicates that there are more results which aren't shown.

felixSchl:
I've applied part of the suggestion, but the problem that I see is that the server still has to do a lot of work when navigating to say page 20 since it needs to `drop` from the result set. I guess we could cache search results in memory for a certain amount of time since last access? But I feel this introduces non-essential complexity.

hdgarrood:
Yes, I don't want to do caching like that, I agree that that could be very complex and not really give us all that much in return. That's why I proposed having hard cut-off point at 10 pages.

felixSchl:
Alright, I have added such a limit

src/Handler/Search.hs
```diff
@@ -29,17 +28,20 @@ import qualified XMLArrows
...
-  page <- maybe 1 (max 1) . ((readMaybe . T.unpack) =<<) <$> lookupGetParam "page"
+  page <- maybe 1 (max 1) . ((readMaybe . unpack) =<<) <$> lookupGetParam "page"
 
   case mquery of
     Nothing -> redirect HomeR
     Just query -> do
       (results, hasMore) <- case tryParseType query of
-        Just ty | not (isSimpleType ty) -> searchForType page ty
-        _ -> searchForName page (toLower query)
+        Just ty | not (isSimpleType ty) -> searchForType (min maxPages page) ty
```
hdgarrood: I think we should extract the part that makes sure `page <= maxPages` into a separate function at the time we take it out of the query string, together with all of the other validations and things we do to it.

src/Handler/Search.hs
```diff
@@ -7,6 +7,10 @@ module Handler.Search
 import Import
 import Data.Trie (elems, submap)
 import Data.Version (showVersion)
+import Data.Maybe (isNothing)
```
hdgarrood: Can you re-export these functions and `MaybeT` from `Import.NoFoundation`? I think they're ubiquitous enough that it's better to do that to help avoid having too many imports here.

src/Handler/Search.hs
```diff
@@ -21,22 +25,52 @@ import qualified Language.PureScript as P
...
+resultsPerPage :: Int
+resultsPerPage = 50
+
+maxPages :: Int
+maxPages = 10
+
 getSearchR :: Handler TypedContent
 getSearchR = do
   mquery <- lookupGetParam "q"
+  page <- maybe 1 (max 1) . ((readMaybe . unpack) =<<) <$> lookupGetParam "page"
```
hdgarrood: I think we should extract all of the parsing and validation logic for the page parameter into a separate function.

src/Handler/Search.hs
```diff
@@ -21,22 +25,52 @@ import qualified Language.PureScript as P
...
   where
-    htmlOutput :: Text -> [SearchResult] -> Handler Html
-    htmlOutput query results = do
+    mkPaginationW :: Int -> Bool -> Handler (Maybe Widget)
+    mkPaginationW page hasMore = runMaybeT $ do
+      cr <- MaybeT getCurrentRoute
+      MaybeT $ do
+        urp <- getUrlRenderParams
+        getParams <- Map.fromList . reqGetParams <$> getRequest
+        let mkPageLink p = urp cr $ Map.toList
```
hdgarrood: For the pagination widget, how about having a link for all pages up to the current one? Also can you explain what the purpose of `getParams` is here?

src/Handler/Search.hs
```diff
@@ -84,15 +118,24 @@ routeResult SearchResult{..} =
   ppkgName = PathPackageName hrPkgName
   pversion = PathVersion hrPkgVersion
 
-searchForName :: Text -> Handler [SearchResult]
-searchForName query = do
+searchForName :: Int -> Text -> Handler ([SearchResult], Bool)
+searchForName page query = do
+  db <- atomically . readTVar =<< (appDatabase <$> getYesod)
+  let xs = map fst
+            $ take (resultsPerPage + 1)
```
hdgarrood: This new logic seems like it should be extracted into a separate function too.

src/Handler/Search.hs
felixSchl: So you mean the pages keep growing as you advance? So on page 4 you would see "1 2 3 [4] next", but on page 3 you would only see "1 2 [3] next"?

hdgarrood: Yes - the idea being, you might think "okay I should have seen the thing I am looking for by now so I don't think looking further will help, but I remember seeing something on page 2 which was close to what I want so maybe the answer is in there somewhere"

hdgarrood: Also because if we're on page 4, we know that pages 1, 2, and 3 exist, but we don't know if 6,7, or 8 exist.

felixSchl:
Alright, I've addressed all of the above

src/Handler/Search.hs
felixSchl: Sorry, forgot to explain `getParams`, it's to modify the `?page=` param for each link

src/Handler/Search.hs
```diff
@@ -52,6 +85,9 @@ getSearchR = do
     isSimpleType P.TypeConstructor{} = True
     isSimpleType _ = False
 
+    lookupCurrentPage :: Handler Int
+    lookupCurrentPage = maybe 1 (max 1) . ((readMaybe . unpack) =<<) <$> lookupGetParam "page"
```
hdgarrood: Could  `(min maxPages page)` go in here too?

src/Handler/Search.hs
```diff
@@ -84,15 +120,21 @@ routeResult SearchResult{..} =
   ppkgName = PathPackageName hrPkgName
   pversion = PathVersion hrPkgVersion
 
-searchForName :: Text -> Handler [SearchResult]
-searchForName query = do
+applyPagination :: Int -> [a] -> ([a], Bool)
+applyPagination page xs =
+    let ys = take (resultsPerPage + 1) $ drop ((page - 1) * resultsPerPage) xs
```
hdgarrood: This will give us `resultsPerPage + 1` results per page, right? That doesn't seem right :smile:

How about this?

```haskell
let
  (ys, more) = splitAt resultsPerPage $ drop ((page - 1) * resultsPerPage) xs
in
  (ys, not (null more))
```

hdgarrood: Also this function could do with a brief comment, I think

src/Handler/Search.hs
```diff
@@ -84,15 +120,21 @@ routeResult SearchResult{..} =
...
+searchForName :: Int -> Text -> Handler ([SearchResult], Bool)
+searchForName page query = do
+  db <- atomically . readTVar =<< (appDatabase <$> getYesod)
+  return $ first (map fst) $ applyPagination page $ concat $ elems $ submap (encodeUtf8 query) db
+
+searchForType :: Int -> P.Type -> Handler ([SearchResult], Bool)
+searchForType page ty = do
   db <- atomically . readTVar =<< (appDatabase <$> getYesod)
-  return (map fst (take 50 (concat (elems (submap (encodeUtf8 query) db)))))
+  return $ first (map fst) $ applyPagination page $ sortBy (comparing snd) (mapMaybe (matches ty) $ concat (elems db))
```
hdgarrood: This line is getting a bit too long now, I think

hdgarrood:
Also yes, let's use a header in the JSON response to indicate whether there are more results.

src/Handler/Search.hs
felixSchl: nice

src/Handler/Search.hs
felixSchl: Sure. I've rewritten it in a more verbose but more readable way.

hdgarrood: :+1:

src/Handler/Search.hs
```diff
@@ -86,7 +86,12 @@ getSearchR = do
     isSimpleType _ = False
 
     lookupCurrentPage :: Handler Int
-    lookupCurrentPage = maybe 1 (max 1) . ((readMaybe . unpack) =<<) <$> lookupGetParam "page"
+    lookupCurrentPage = fmap (fromMaybe 1) <$>
+        runMaybeT $ do
```
hdgarrood: I think this would be slightly clearer without `MaybeT`, actually:

```haskell
do
  mpage <- lookupGetParam "page"
  return $ case mpage >>= (readMaybe . unpack) of
    Just page -> max 1 $ min maxPages page
    Nothing -> 1

**Comment by felixSchl** [comment]
Alright, I've removed the use of `MaybeT` and am now rendering the `Link` header (I took inspiration from: https://developer.github.com/v3/guides/traversing-with-pagination/#navigating-through-the-pages). Sorry for the amount of changed code in latest commit, but there was a bit of common logic between rendering the header and the widget, so I extracted

File: src/Handler/Search.hs [hunk 10]
```diff
@@ -21,26 +24,87 @@ import qualified Language.PureScript as P
...
-    htmlOutput query results = do
+    getPageLinkRenderer :: Handler (Maybe (Int -> Text))
+    getPageLinkRenderer = runMaybeT $ do
+      cr <- MaybeT getCurrentRoute
+      MaybeT $ do
+        urp <- getUrlRenderParams
+        getParams <- Map.fromList . reqGetParams <$> getRequest
+        return $ Just $ \p ->
+          urp cr $ Map.toList
+                 $ Map.insert "page" (pack $ show p) getParams
```
hdgarrood: Come to think of it I'm not sure `MaybeT` is really adding much here either.

static/css/pursuit.css
```diff
@@ -619,6 +619,47 @@ ol li {
...
+.pagination__top {
+    padding-bottom: 1.5em;
+}
+
+.pagination__bottom {
+    padding-top: 1.5em;
+}
+
+.pagination {
+    position: relative;
```
hdgarrood: Why are the `position` rules needed?

src/Handler/Search.hs
```diff
@@ -21,26 +24,87 @@ import qualified Language.PureScript as P
...
+
+    mkPaginationW :: Int -> Bool -> Handler (Maybe Widget)
+    mkPaginationW page hasMore = getPageLinkRenderer >>= \mr -> case mr of
+        Nothing -> return Nothing
+        Just mkPageLink ->
+          let allPages = [1..page]
+              mmNextPageLink = if hasMore
+                                  then Just $ if page >= maxPages
```
felixSchl: I think for a one off job like this the cognitive burden of `Maybe (Maybe a)` is less than introducing another ADT the reader of the code has to keep in memory (and then forget about after realizing how irrelevant it is)

src/Handler/Search.hs
```diff
@@ -21,26 +24,87 @@ import qualified Language.PureScript as P
...
       fr <- getFragmentRender
+      mPaginationW <- mkPaginationW page hasMore
       content <- defaultLayout $(widgetFile "search")
       sendResponseStatus ok200 content
 
-    jsonOutput = fmap toJSON . traverse searchResultToJSON
+    jsonOutput :: [SearchResult] -> Int -> Bool -> Handler Value
+    jsonOutput results page hasMore = do
+        when (hasMore && page <= maxPages) $ void $ do
+          mkPaginationHeader page hasMore >>= \mHeader -> case mHeader of
```
felixSchl: I didn't want to add any syntax extensions, but I think `LambdaCase` would be pretty cool for all these cases:

```
mkPaginationHeader page hasMore >>= case
  Nothing -> pure ()
  Just ...
```

felixSchl: Oh, hold on, it's already enabled, will transform

static/css/pursuit.css
felixSchl: I am no CSS wiz, I'll try to remove

felixSchl:
Alright, I've addressed all the things

src/Handler/Search.hs
hdgarrood: In the past I would have agreed with you, but every single time I've used `Maybe (Maybe a)` like this, I've regretted it. In my opinion it's a far larger cognitive burden to have to remember what the meaning of `Nothing` vs `Just Nothing` is.

However before you address that, I noticed that this is what page 10 now looks like on my machine:
screenshot from 2017-07-15 13-28-42

and I was thinking maybe it would make more sense to have the "further results have been omitted" message after the last search result rather than inside the pagination widget.

Also I think the clickable regions in the "goto page n" links should still be made a little larger. Also I don't think there should be non-clickable gaps in between the links. It might also be nice to have the background change colour slightly on hover so that it's clear what you're clicking on. Finally I think it might be slightly nicer to have the next page link text include a chevron character like "next page &raquo;" and similar for the prev page link?

templates/search.hamlet
```diff
@@ -44,3 +48,12 @@
           <span .result__actions__item>
             <span .badge.badge--module title="Module">M
             #{moduleName}
+
+  $if hasOmittedResults
+    <hr />
+    <div .result__omission>
+      More results have been omitteded
```
hdgarrood: I think "Further results have been omitted" would be better here.

static/css/pursuit.css
```diff
@@ -619,6 +619,59 @@ ol li {
 .result__actions__item + .result__actions__item {
   margin-left: 0.65em;
 }
+.result__omission {
```
hdgarrood: I just need to work out how we will deal with this CSS, because it's generated from a LESS file (in this repo), but there's also a copy of it in the PureScript repo, and they're starting to diverge.

src/Handler/Search.hs
```diff
@@ -21,26 +23,87 @@ import qualified Language.PureScript as P
 
 import qualified XMLArrows
 
+resultsPerPage :: Int
```
hdgarrood: Was this 50 originally?

static/css/pursuit.css
hdgarrood: https://github.com/purescript/purescript/issues/3016 if you're wondering what's going on

PR #319, opened 2017-07-14
Fix source links for re-exports

### LINKED ISSUE #304: Source links don't work on functions from re-exported modules from other packages

felixSchl:
Fixes #304 

There's a potentially blocking caveat to this solution that I have not found a way around: In order to render source links for re-exports, we need to load the `VerifiedPackage` for every dependency. And since the `packageAsHtml` is pure, we have to create a lookup map prematurely, rather than on demand. Would be simple join to get this info if we had a database out the back, but now it's one file read per dependency.

### DIFF
--- a/src/Import/NoFoundation.hs
+++ b/src/Import/NoFoundation.hs
@@ -1,5 +1,6 @@
 module Import.NoFoundation
     ( module Import
+    , hush
     ) where
 
 import ClassyPrelude.Yesod   as Import
@@ -7,3 +8,6 @@ import Settings              as Import
 import Yesod.Core.Types      as Import (loggerSet)
 import Yesod.Default.Config2 as Import
 import Control.Category      as Import ((>>>), (<<<))
+
+hush :: Either a b -> Maybe b
+hush = either (const Nothing) Just
--- a/src/TemplateHelpers.hs
+++ b/src/TemplateHelpers.hs
@@ -5,6 +5,7 @@ import qualified Data.List.Split as List
 import qualified Data.Map as Map
 import Data.Traversable (for)
 import Data.Version (Version)
+import Data.List (nub)
 import Data.Text (splitOn)
 import Data.Time.Format as TimeFormat
 import Text.Blaze.Html5 as H hiding (map, link)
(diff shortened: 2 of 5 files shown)


src/TemplateHelpers.hs
```diff
@@ -125,7 +128,18 @@ renderReadme = \case
 renderHtmlDocs :: D.VerifiedPackage -> Text -> Handler (Maybe Html)
 renderHtmlDocs pkg mnString = do
   htmlRenderContext <- getHtmlRenderContext
-  let docsOutput = packageAsHtml (htmlRenderContext pkg) pkg
+
+  depHtmlRenderContexts <- do
```
hdgarrood: I found this a little difficult to follow, keeping track of how many arguments have been applied to the `htmlRenderContext` function at a given point. How about extracting a function `collectDepsHtmlRenderContexts :: Package a -> Handler (Bower.PackageName -> P.ModuleName -> Maybe D.HtmlRenderContext)`?

hdgarrood: Also I think it would be good to split out some parts of this into a `where` clause e.g. something along the lines of `makeLookupEntry = fmap ((pkgName,) . htmlRenderContext) . either (const Nothing) Just`

felixSchl: Hm, how about we just move the entire `deplHtmlRenderContexts` into a where clause? This way you get type annotations to indicate how far the function has been applied and all complexity is localized. What do you think?

src/TemplateHelpers.hs
```diff
@@ -125,7 +128,18 @@ renderReadme = \case
...
+    Map.fromList . catMaybes <$> do
+      for (D.pkgResolvedDependencies pkg) $ \(pkgName, version) -> do
+        fmap (pkgName, ) <$> do
+          either (const Nothing) Just <$> do
+            fmap htmlRenderContext <$> do
+                lookupPackage pkgName version
+
+  let docsOutput = flip packageAsHtml pkg $ \case
+        D.Local m -> Just $ htmlRenderContext pkg m
+        D.FromDep pkgName m -> return m <**> Map.lookup pkgName depHtmlRenderContexts
```
felixSchl: I looked for a function on hoogle `a -> f (a -> b) -> f b`, but only could find `f a -> f (a -> b) -> f b`, hence the `return m` here. Do you know of such a function? Maybe I missed it.

felixSchl:
With respect to performance, do you think we should parallelise the file reads? Also maybe there's a way to per-emptively discard dependencies we are not re-exporting anything from.

hdgarrood:
Reading the files concurrently would be good if you can but I'm not sure if `Handler` can cope with that sort of thing. You could also look through all of the package's re-exports first to get a list of only the ones you need.

src/TemplateHelpers.hs
hdgarrood: I'd probably just use `fa2b <*> pure a` or `fmap ($ a) fa2b`.

src/TemplateHelpers.hs
hdgarrood: I think it's simpler to split off a top-level function actually, because then when you're reading the code you can immediately know that this function doesn't depend on extra stuff in scope inside the other function, such as bound variables for its arguments and other stuff in `where` clauses.

src/Import/NoFoundation.hs
```diff
@@ -1,9 +1,13 @@
...
+    , hush
     ) where
 
 import ClassyPrelude.Yesod   as Import
 import Settings              as Import
 import Yesod.Core.Types      as Import (loggerSet)
 import Yesod.Default.Config2 as Import
 import Control.Category      as Import ((>>>), (<<<))
+
+hush :: Either a b -> Maybe b
```
hdgarrood: :+1:

src/TemplateHelpers.hs
```diff
@@ -184,6 +178,26 @@ docLinkRoute D.LinksContext{..} srcModule link = case D.linkLocation link of
       (PathVersion version)
       (P.runModuleName modName)
 
+getDepHtmlRenderContexts
+    :: D.Package a
+    -> Handler (Map Bower.PackageName (P.ModuleName -> HtmlRenderContext))
```
hdgarrood: I think it might be a bit neater to move the `Map.lookup` inside here so that this function returns a `Handler (Bower.PackageName -> P.ModuleName -> Maybe HtmlRenderContext`

felixSchl:
Alright, I've made those changes. I think the function is fairly easy to follow now. I've added a helper function `hush` to the import module for `Either a b -> Maybe b`. It's also doing less now by scanning for re-exports first. With respect to reading concurrently, would a primitive forkIO /mvar solution do the trick here or add too much noise?

hdgarrood:
Merged in #345. Thanks!

Issue #150, opened 2015-08-27
Remove module lists from README files


garyb:
I remember talking about this somewhere so apologies if it's a duplicate - I couldn't see one anywhere.

I suggest an easy quick-fix for this would be to remove sections in the markdown that start with the title `Module documentation`.

paf31:
Often, I want to include different information on Pursuit. What about adding `PURSUIT.md`, and falling back to `README.md` with no `Module Documentation` section if it's missing?

garyb:
Oh! That's right, we did talk about that. +:100: for that instead.

hdgarrood:
What different information do you want to include? Is it just the "module documentation" part that would be different?

I think I'd rather we remove "module documentation" parts from READMEs entirely and stop checking markdown documentation into repositories, because this encourages people to read docs on GitHub.

garyb:
I was thinking about that as an option too, but then I'd probably want to link to Pursuit in the README, so you'd end up with a recursive link on the Pursuit page for the module :smile: (of course we could eliminate that too).

hdgarrood:
I don't see that as a problem - I was assuming that most people would link to Pursuit in their readmes, probably using one of those badges like Travis has. Also, I'd rather just leave it there - I don't think it does any harm, and I want to keep this kind of fiddling to an absolute minimum.

garyb:
Well, I'd like to include it as a sentence or something, unless people know what "pursuit" means it doesn't really shout "documentation here!".

hdgarrood:
True, that does sound sensible. Perhaps we could also start putting Pursuit URLs into the repository website field that GitHub provides.

aspidites:
My initial reaction was that having yet another file to manage would feel a bit redundant (even if we say that we fallback to README.md). As I think more about it, I think that having a separate PURSUIT.md wouldn't be much different than Python projects that have entire directories for sphinx to parse in addition to the inline documentation. 

My only remaining concern (if I ever get to a point where I'm writing library documentation) would be how much overlap there might be in having both files (eg. maybe something that's in PURSUIT.md is worth repeating in README.rst).

As another idea, would it be worth having special comments that are parsed by pursuit? Something like:

```markdown
<!-- Start Pursuit -->
# Introduction
This is a simple introduction.
<!-- End Pursuit -->
# Details
This is more detail that belongs in a README but not [Pursuit](pursuit.purescript.org).
```

[...]

PR #322, opened 2017-07-29
Add margins to highlighted code blocks

owickstrom:
This PR adds a top and bottom margin to `.highlight` blocks, with the same margin size as paragraphs. Without it, adjacent code blocks look like a single code block (in the example below this becomes confusing).

Note that I haven't set up Pursuit on my machine fully, and thus couldn't test this styling locally. I've just tested it in the Chrome inspector before making the changes, so someone might want to run this locally and verify on a few other READMEs.

## Before

image

## After

image


hdgarrood:
I agree that I'd like to fix this, but I'm a little apprehensive about doing it like this because:

* If the same issue exists in API docs, i.e. for adjacent code blocks in doc-comments, I'd like to fix it there too
* I want to stop using the GitHub API for rendering readmes, and IIRC it's because of the GitHub API that these code blocks have a `highlight` class.

Did you have difficulty getting Pursuit working locally? I'd prefer this were more thoroughly tested first too.

owickstrom:
I see!

I'll first try to get Pursuit up and running locally. The reason I didn't before was just lack of time, I squeezed this in before work. Then perhaps I might find some time to try an alternative to the GitHub API rendering. Do you have any other recommendations on what could be used? Maybe Pandoc?

hdgarrood:
We are already using Cheapskate for doc-comments, so that would be the obvious choice. However, for the time being, the only way of getting the readme will be via the GitHub API, as currently it's not included in the json produced by purs publish. For now we could probably change the GitHub API call to ask for the raw README rather than rendered HTML, and then render it ourselves using Cheapskate. We would also have to work out how to handle the case where someone's README isn't Markdown.

hdgarrood:
You might want to take a look at the discussion in https://github.com/purescript/pursuit/issues/150 for context, as well.

hdgarrood:
I think I'll close this for now since there's been no movement for a while, but in general I'm happy to discuss changes like this (provided the above questions I raised have been addressed). Also, just FYI, I'm planning to change Pursuit soon so that it uses the CSS from the purescript library, so changes like this might need to be made there rather than here in the future.

Issue #291, opened 2017-03-20
Link to Prim from the front page


Andrew-Webb:
If I go to Pursuit's front page and do a browser find (ctrl+f) for "prim", there is no match.

If I search (Pursuit search, not browser find) for something that I know exists in the https://pursuit.purescript.org/builtins/docs/Prim page, then I don't get results from `module Prim`.

E.g. if I search for "Record", I get results only for modules `Data.Either.Compact` and `Data.RateLimit`.

As a result, I keep https://pursuit.purescript.org/builtins/docs/Prim bookmarked separately from Pursuit, since it's such a useful page.  But it would be nice if it was reachable by the normal Pursuit user interface.

hdgarrood:
We have an issue for the search results already: #265. I'm not sure about the front page; the front page is a list of packages, and Prim is a module which is built in to the compiler so it doesn't exist in any package, and so it's not immediately clear how to include it. I'll think about it though.

Andrew-Webb:
> Prim is a module which is built in to the compiler so it doesn't exist in any package, and so it's not immediately clear how to include it

https://github.com/hdgarrood/purescript-prim refers to `purescript-prim` as a pseudo-package, so maybe it could be listed on the front page between `purescript-prelude` and `purescript-prng`.  Kinda nice that it would be situated so closely to the prelude.

hdgarrood:
I should take that down, it's not used any more.

hdgarrood:
Fixed by #331.

Issue #326, opened 2017-08-23
Page listing all core libraries


hdgarrood:
I think it could be helpful to put all the core libraries on their own page on Pursuit, so that people who are relatively new to PureScript can look through it and familiarise themselves with them. I think this would also be a good place to clarify what a "core library" is, as this doesn't seem to be particularly clear to newcomers right now. I'd want to replace the link to Prelude on the front page with a link to this core libraries page (of course, Prelude would then be listed on that page).

I think this would go a long way toward addressing the issue raised in #158 and also some of the issues raised in https://github.com/purescript/documentation/issues/43.

Thimoteus:
What I've wanted for a while is for pursuit to order first by core libraries, then third party libraries. A lot of names are used across libs and I almost always want the one that defined in a core repo.

hdgarrood:
Do you mean when searching by name?

Thimoteus:
Yes, for example: [apply](https://pursuit.purescript.org/search?q=apply), it's not until the 4th listing that I get the definition of `<*>`

hdgarrood:
Right, agreed. I'd prefer to consider that in a separate issue though.

Thimoteus:
Understood.

PR #331, opened 2017-08-27
Add Prim to homepage

joneshf:
N.B. I haven't tried to run this. But I wanted to throw this out for discussion/opinion.

Should address #291.

Currently, there's no link at all to this page on Pursuit, any other PureScript website, or any other PureScript GitHub repo. What that means is that unless you're in the know, you can't find the page. I get that search results would be really great to have, but that shouldn't prevent anyone in the community from finding these docs.

I'm not saying this is the greatest thing ever, and I know it can be improved. But, this gets us from zero to nonzero, which is more valuable than being stuck trying to go from zero to ten.

### DIFF
--- a/templates/homepage.hamlet
+++ b/templates/homepage.hamlet
@@ -8,7 +8,7 @@
     Pursuit hosts API documentation for <a href="http://www.purescript.org/">PureScript</a> packages. It lets you search by package, module, and function names, as well as approximate type signatures.
 
   <p>
-    To get started, have a look around the <a href="/packages/purescript-prelude">Prelude</a>, or try one of these example searches:
+    To get started, have a look around <a href=@{BuiltinDocsR "Prim"}>Prim</a>, the <a href="/packages/purescript-prelude">Prelude</a>, or try one of these example searches:
 
   <ul .ul--search>
     <li><a href="/search?q=map">map</a></li>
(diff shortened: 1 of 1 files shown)


hdgarrood:
> Currently, there's no link at all to this page on Pursuit, any other PureScript website, or any other PureScript GitHub repo

This isn't actually true: whenever any `Prim` type appears in a type signature, it is rendered as a link to this module. Try clicking on any occurrence of `Int`, `String`, `Array` etc in <https://pursuit.purescript.org/packages/purescript-prelude/3.1.0/docs/Prelude>, for instance.

Anyway I think it would make the most sense to put this link in the same sentence as the one that suggests looking at the Prelude, which appears slightly earlier in the HTML.

joneshf:
Moved to the same sentence.

hdgarrood:
Looks good, thanks! This is definitely an improvement, although I think we could take this further still by explaining what Prim contains, so that people who don’t already know what Prim is are more likely to be able to benefit from it. At that point I think it’s likely to get a bit too wordy for the front page though, so I’d suggest we put together a separate page for this. This page could give you an overview and suggested perusal order for the core libraries, too: see #326.

Anyway I’m going to go ahead and merge this now, we can do all that later.

joneshf:
Yeah, makes sense. Thanks!

joneshf:
What do we need to do to update https://pursuit.purescript.org/ to show this?

hdgarrood:
I can’t remember the exact details why this is the case but we need to wait until 0.12 of the compiler is out.

hdgarrood:
(We could also backport but I’ve generally not bothered with backporting things because it takes up time that I don’t really have nowadays.)

joneshf:
@hdgarrood Did 0.11.7 fix whatever the problem was? If so, can we update the site?

hdgarrood:
The issue was that lots of links for types rendered in documentation were broken. I think it should be fixed in 0.11.7; when I get a moment, I'll try updating Pursuit to use 0.11.7 and check. Thanks for the reminder :)

PR #336, opened 2017-09-24
Upgrade to PureScript 0.11.6.

chexxor:
I'd like to fix the incorrect type signature of RowCons, so I think we need to upgrade the version of the PureScript library here.

Looks like it was fixed awhile in the PS repo, so just need to update our dep here: https://github.com/purescript/purescript/issues/2920

idk how updating the server instance works, but this is a good first step.

### DIFF
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -1,5 +1,5 @@
 name:              pursuit
-version:           0.6.1
+version:           0.6.2
 cabal-version:     >= 1.8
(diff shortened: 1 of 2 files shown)


hdgarrood:
I think it would be best to wait until 0.12, because I introduced a bug in docs generation in 0.11.6 which breaks most of the links in docs. It’s fixed in master now so all we need to do is wait.

Issue #268, opened 2017-01-06
Add links for packages and modules on search results


soupi:
Search results currently display the package and the module result comes from, but those aren't clickable links. Maybe make those link to the package/module?

PR #341, opened 2017-12-06
Add RSS feed for latest packages

joncfoo:
Hi, I use RSS quite a bit and thought Pursuit ought to have a feed for the latest uploaded packages - I'd like my feed reader to tell me of updates :)

This PR adds a RSS feed at `/latest-packages` which lists the same packages that are displayed on the homepage.  It uses the `yesod-newsfeed` library which integrates well with the existing Yesod app.  There are two outstanding things that I've noted below:

1. Where should the link to the RSS feed go?
    - Ideally it should be embedded in the `head` tag on the homepage and also as an icon on the homepage somewhere.  Is there a way to conditionally include the link in `head` only if the layout template is processing on the homepage?
2. Caching the RSS feed
    - It's probably a good idea to cache the feed but I'm unsure how to go about doing so.  I had a look at `src/Handler/Caching.hs` but wasn't sure how to go about turning the `TypedContent` feed into a ByteString.  I'm unfamiliar with Yesod but happy to dig in if given some pointers.

### DIFF
--- a/config/routes
+++ b/config/routes
@@ -12,6 +12,7 @@
 !/packages/#PathPackageName/docs/#Text                 PackageModuleDocsR          GET
 !/packages/#PathPackageName/badge                      PackageBadgeR               GET
 !/packages/#PathPackageName/available-versions         PackageAvailableVersionsR   GET
+/packages-latest                                       PackagesLatestR             GET
 
 /builtins/docs/#Text                                   BuiltinDocsR                GET
 
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -69,6 +69,7 @@ library
                  , yesod-core                    >= 1.4.6      && < 1.5
                  , yesod-static                  >= 1.4.0.3    && < 1.6
                  , yesod-form                    >= 1.4.0      && < 1.5
+                 , yesod-newsfeed
                  , classy-prelude                >= 0.10.2
                  , classy-prelude-conduit        >= 0.10.2
                  , classy-prelude-yesod          >= 0.10.2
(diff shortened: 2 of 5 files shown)


paf31:
I do like the idea of being able to send package releases to the `purescript` Twitter feed 😄

hdgarrood:
Yeah, this seems like a good addition :) there's just a couple of things I want to sort out before I get to this, like the other open PRs.

joncfoo:
@hdgarrood I'd be happy to help out with other PRs - if you have something in mind that someone relatively new can take on do let me know

hdgarrood:
The way the cache works currently is that the Yesod app writes the response to a specific file path just before returning it to the client, and then for subsequent requests, nginx is configured to just serve that file if it exists, bypassing the Yesod app entirely. The Yesod app also invalidates the cache appropriately when new versions of packages are uploaded, by deleting the relevant files.

I wonder if a better approach might be to use nginx's directives like [`proxy_cache_path`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_path).

hdgarrood:
Come to think of it, it's good that you're interested in nixifying the community server, because any changes to the caching will need careful testing together with nginx; that will be much easier if I can get it running locally.

hdgarrood:
@joncfoo If you'd like to take something on, how about #268?

joncfoo:
sounds good - can you assign it to me?

hdgarrood:
I cannot unfortunately, I think you need to either be the issue author or have commit access to the repo be an assignee. Feel free to comment on it to say you're working on it if you want, though.

PR #346, opened 2017-12-21
Include child declaration types in search results

### LINKED ISSUE #264: Type signatures for class members are missing in search results
Not sure if there's already an issue for this - I couldn't see one anywhere, which I find surprising since this hasn't worked for a while!

Example: https://pursuit.purescript.org/search?q=traverse

Note the result for the `Traversable` `traverse`:

```
traverse

Data.Traversable (purescript-foldable-traversable)
```

As compared with a non-class-function:

```
traverse :: forall a w. Ord a => a -> Graph a w -> List a

Data.Graph (purescript-digraph)
```

hdgarrood:
Fixes #264. Also fixes the same issue for data constructors.

### Before
screenshot from 2017-12-21 17-25-13
screenshot from 2017-12-21 17-24-59
### After
screenshot from 2017-12-21 17-25-54
screenshot from 2017-12-21 17-25-38

### DIFF
--- a/src/Handler/Database.hs
+++ b/src/Handler/Database.hs
@@ -107,23 +107,66 @@ createDatabase = do
                , ( SearchResult (bowerName pkgMeta)
                                 pkgVersion
                                 (fromMaybe "" declComments)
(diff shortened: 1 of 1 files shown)


garyb:
Awww yis!

PR #350, opened 2017-12-24
Include Prim entries in search results (fixes #265)

### LINKED ISSUE #265: Prim declarations do not show up in search results
Try searching for `Int` or `String`, for instance.

hdgarrood:
I've tried to avoid breaking the JSON search results format as much as possible, but this could cause problems for IDE plugins. Here's what a `Prim` result looks like:

```
{
  "text":"<p>A 32-bit signed integer. See the purescript-integers package for details\nof how this is accomplished when compiling to JavaScript.</p>\n<p>Construct values of this type with literals:</p>\n<pre><code>x = 23 :: Int\n</code></pre>\n",
  "markup":"<p>A 32-bit signed integer. See the purescript-integers package for details\nof how this is accomplished when compiling to JavaScript.</p>\n<p>Construct values of this type with literals:</p>\n<pre><code>x = 23 :: Int\n</code></pre>\n",
  "url":"http://localhost:3000/builtins/docs/Prim#t:Int",
  "version":"0.11.7",
  "package":"<builtin>",
  "info":{
   "typeOrValue":"TypeLevel",
    "module":"Prim",
    "typeText":null,
    "title":"Int",
    "type":"declaration"
  }
}
```

/cc @kRITZCREEK @nwolverson @FrigoEU @coot (have I missed any ide plugin maintainers here?)

### DIFF
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -56,14 +56,21 @@ instance PathPiece VerificationKey where
 
 -- | A single search result.
 data SearchResult = SearchResult
-  { hrPkgName    :: PackageName
-  , hrPkgVersion :: Version
-  , hrComments   :: Text
-  , hrInfo       :: SearchResultInfo
+  { srSource   :: SearchResultSource
+  , srComments :: Text
+  , srInfo     :: SearchResultInfo
   }
   deriving (Show, Eq, Generic)
 
 instance NFData SearchResult
+
+-- | Tells you where a search result came from.
+data SearchResultSource
+  = SourceBuiltin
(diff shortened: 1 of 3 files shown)


nwolverson:
What's the potential breakage?  looks from glance like the existing format, is it the `<builtin>` package name you refer to?

hdgarrood:
Yeah, so that's not a valid package name, and also I've put the compiler version in the "version" field rather than the package's version. Other than that it's the same. You might not need to do anything but I just thought I'd check with you all.

PR #357, opened 2017-12-27
New help pages

grossbart:
@hdgarrood I reviewed your commit 92463dd9b1b3230614ca0b5d01279980dc86d384 and like it a lot. The explanations on searching are clear to me (still a beginner) and the split into user and author help pages is a good idea to provide help for the context that it is needed in.

I added three small commits that all contain a small description in their respective commit message.

1. Move anchors to keep title in viewport
2. Structure help menu with existing styles
3. Use proper typographic quotes

I want to point out that number 3 could be controversial as described in its commit message and you might not want to merge that one (although it would look better).

As for number 2: this is a suggestion of how to use the existing styles better (see screenshot), but if you would want to have a single list that looks more like an index, we'd have to introduce that style (because the spacing in the way you used it before wasn't working so well, I think). To me, number 2 is acceptable and would maybe or maybe not have to be revisited once more topics are added. (In which case you would need to help me with developing the styles in purescript/purescript and then integrate them into purescript/pursuit, I didn't see an easy way right away).

pursuit-help

### DIFF
--- a/src/Handler/Help.hs
+++ b/src/Handler/Help.hs
@@ -36,25 +36,25 @@ helpLayout forWhom inner =
     <div .col.col--aside>
       <dl .grouped-list>
         <dt .grouped-list__title>
-          Contents
+          Overview
         <dd .grouped-list__item>
           <a href="/help">Help index
+      <dl .grouped-list>
+        <dt .grouped-list__title>
+          Pursuit users
         <dd .grouped-list__item>
-          <a href="/help/users">Pursuit users
-          <ul>
-            <li>
-              <a href="/help/users#searching">
-                Searching
+          <a href="/help/users#searching">
+            Searching
+      <dl .grouped-list>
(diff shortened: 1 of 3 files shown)


hdgarrood:
Looks fantastic, thank you! As you say, I think we're likely to end up with a mix of straight and proper quotes if we pass them through as-is, so I think I'd rather not. I would like to investigate a smart quote filter though; would you like to take a look at that or should I?

hdgarrood:
I've merged your first two commits manually so I'll close this for now; we can deal with smart quotes separately.

grossbart:
Thanks for opening up a separate issue on the smart quotes problem, that makes total sense. If no one before me tackles it, I will give it a try, but don't expect it anytime soon 😬

hdgarrood:
Of course :+1:

PR #366, opened 2018-01-14
Fix the package list sorting

paf31:
Sorting by just package name doesn't work, we need to normalize the names first.

_Note_: I haven't tested this yet, sorry.


paf31:
I'll try to get this working first, then reopen this.

Issue #292, opened 2017-03-24
show dependents


matthewleon:
It can be very handy to see a package's dependents to:

1) See examples of how the package is used.
2) Provide a rough heuristic for the package's popularity in the ecosystem.

Thoughts?

hdgarrood:
Sounds good to me.

matthewleon:
Adding a note here that `psc-package` makes this much more easily achievable: https://github.com/purescript/package-sets/blob/master/packages.json

hdgarrood:
I don't think I'd be comfortable with using the official package set to obtain this information, since many packages aren't in there.

LiamGoodacre:
Related to https://github.com/purescript/pursuit/issues/141, although, the other way around.

kurtmilam:
I think this would be a huge help for anyone trying to learn PureScript. It's almost impossible to find practical implementation examples of some libraries I'd also like to highlight GitHub's facilities for searching globally within code. After spending several hours searching for solutions and poking at getting `purescript-quickcheck-laws` working in a learning project, I (re)acquainted myself with this GitHub feature.

Seaching for `in:file language:purescript "purescript-quickcheck-laws` returned 194 results in GitHub repositories, and I had a working implementation within a quarter of an hour.

![image](https://user-images.githubusercontent.com/385695/58385886-be6d6200-7ff7-11e9-9301-608698f3dbbc.png)

I'm motivated to learn PureScript, but it's frustrating to spend hours searching for simple implementation examples repeatedly.

hdgarrood:
I would be happy to leverage GitHub search, but that's a different feature request really: #168.

On reflection I'm a little less keen on implementing this as it seems difficult from an architectural perspective; see #375.

milesfrain:
Found this issue while trying to figure out what other packages use `purescript-gen` besides `purescript-quickcheck`.
Seems like this is a task for the registry. Discussed a bit in https://github.com/purescript/registry/issues/27
Maybe we could eventually get packages listed in the "Used By" section in github. They currently support these [package managers](https://docs.github.com/en/github/visualizing-repository-data-with-graphs/about-the-dependency-graph#supported-package-ecosystems) - also mentioned in https://github.com/purescript-contrib/setup-purescript/pull/2#issuecomment-671673474

PR #375, opened 2018-05-25
DO NOT MERGE: proposal for adding sqlite for caching dependency relations?

AdamSaleh:
Hi!

First, a disclaimer. This PR probably is quite ugly and unfinished piece of code that I don't expect in any form to be merged. I started working on it ~a week ago, after I tried naively implement reverse-dependency listing and realized that for site to be responsive there would need to be some sort of cache of the package-relation.

If i.e. @hdgarrood would think this approach might have some merit, I would really like to work with somebody to finish this into an actually merge-able piece of code.

If it would be too much work, I have no problem closing this PR and let someone else take a stab at #292 

So the approach in this code is:
* bring in Persistent Sqlite as a database backend
* on the start of the application, load all of the packag versions into Persistent
* On Package display, run an Esqueletto query to get the reverse deps

Right now:
* it adds ~3 minutes to startup
* there are no database updates on create/remove/update of the package (as they should)
* the dependent packages don't take version-ranges into account (as they should)
* the dependent packages are not cross-linked (as they should)
* there are some hard-coded values
* there are some copy-pasted util functions
* I am really unsure about the code-(style/quality), but at least it did compile on my machine :smile: 

So, what do you think?

### DIFF
--- a/config/models
+++ b/config/models
@@ -1,0 +1,8 @@
+-- By default this file is used by `persistFileWith` in Model.hs (which is imported by Foundation.hs)
+-- Syntax for this file here: https://github.com/yesodweb/persistent/blob/master/docs/Persistent-entity-syntax.md
+
+Package
+    dependent Text
+    version Text
+    dependency Text
+    versionRange Text
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -1,5 +1,5 @@
 name:              pursuit
-version:           0.6.3
+version:           0.6.4
 cabal-version:     >= 1.8
 build-type:        Simple
 license:           MIT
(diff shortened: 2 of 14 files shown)


hdgarrood:
Hi! Really sorry I took so long to get around to responding to this - I was right in the middle of my undergrad final year exams back in May which is why this slipped under my radar.

If we are going to use a proper database, I'd rather go the whole hog and use that database for all of the data we're storing; I think that having a mixture of JSON files on the hard disk and data in a database would be awkward. Note that we do already take a note of how many reverse dependencies a package has when building the search index; that might potentially be a good place to store this information.

If this adds 3 minutes to startup time, though, I'm tempted to say we should leave it. Pursuit was always intended to be just a documentation hosting service, not a package registry - and in fact it most likely won't ever become a package registry, in the sense that package managers don't query Pursuit to find install plans or download packages, and probably never will - and I think this feature is really something that a package registry should be taking responsibility for, ideally.

PR #396, opened 2019-07-13
Treat variables in type queries same as type wildcards

### LINKED ISSUE #395: `a -> HTMLElement` and `_ -> HTMLElement` have different results
Going by the [help page for Pursuit](https://pursuit.purescript.org/help/users#searching) my best guess for finding functions that return something of type `HTMLElement` was `HTMLElement` or `a -> HTMLElement`. Neither gives any results although `_ -> HTMLElement` does. It is not clear why.

hdgarrood suggested they should perhaps return the same results. For me being able to find a written suggestion to try `_` vs. `a` could also be enough.

Thanks for this awesome docs service 🙏 !

klntsky:
fix #395

### DIFF
--- a/src/SearchIndex.hs
+++ b/src/SearchIndex.hs
@@ -398,6 +398,7 @@ compareTypes type1 type2 =
   go :: D.Type' -> D.Type' -> WriterT [(Text, Text)] Maybe Int
   go (P.TypeVar _ v1) (P.TypeVar _ v2) = tell [(v1, v2)] *> pure 0
   go t (P.TypeVar _ _) = pure (1 + typeComplexity t)
+  go (P.TypeVar _ _) t = pure (typeComplexity t)
   go (P.TypeLevelString _ s1) (P.TypeLevelString _ s2) | s1 == s2 = pure 0
   go (P.TypeWildcard _ _) t = pure (typeComplexity t)
   go (P.TypeConstructor _ q1) (P.TypeConstructor _ q2) | compareQual q1 q2 = pure 0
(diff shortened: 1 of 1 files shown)


hdgarrood:
This change invalidates the above comment, right? Although perhaps the fact that #395 was opened in the first place suggests that having a query of `a` not match the type `Int` at all wasn't quite the right approach. How about considering it a match, but applying a larger penalty in this case? Say, 2?

hdgarrood:
Hm, I suppose the current handling of `TypeWildcard` doesn't really match up with what the comment says.

klntsky:
> Hm, I suppose the current handling of TypeWildcard doesn't really match up with what the comment says.

I believe the initial idea was that TypeWildcard in a query is a special case and should match any type without additional penalty aside of `typeComplexity` (`typeComplexity` is probably here to ensure that less hairy types appear higher in the results).

> How about considering it a match, but applying a larger penalty in this case? Say, 2?

A larger penalty leads to this:

2019-07-13-213839_678x937_scrot.

IMO the penalty for instantiation should be lower than the penalty for generalization. What about multiplying everything by 2 and charging just 1 for instantiation?

klntsky:
One possible solution is to return negative numbers for exact matches of type constructors:

```
  go (P.TypeConstructor _ q1) (P.TypeConstructor _ q2) | compareQual q1 q2 = pure (-1)
```

klntsky:
As for now I updated the comment to reflect the intention behind the proposed fix (though the best way to resolve the issue is not clear IMO).

PR #408, opened 2020-04-13
Single-command deploys

hdgarrood:
This commit tweaks the deploy scripts so that once a tagged version of
Pursuit has been built on Travis CI and uploaded to GH releases as a
tar.gz bundle, it can be deployed simply by running e.g.

    $ deploy/run.sh v0.7.3

at the command line. This brings us one step closer to auto-deployment.

Specifically, this commit does the following:

- renames the script deploy/run.sh to deploy/remote.sh to better
  indicate that it is intended to be run on the remote server
- updates the deploy script and the systemd service file so that the
  pursuit binary is installed to /usr/local/bin and run from there.
  This is necessary because `tar` refuses to write to a running
  executable file when extracting; we have to use `install` instead.
- clarifies what the deploy script is and isn't responsible for in
  comments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ data
 static/tmp/
 license-generator/tmp/
 /.stack-work/
+bundle/
+pursuit.tar.gz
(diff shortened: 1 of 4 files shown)


hdgarrood:
Thanks!

PR #411, opened 2020-07-06
Add `Event -> _` example

milesfrain:
Using `_` in searches was news to many of us in today's meeting. It produces different results than `Event -> a`.


hdgarrood:
I don't really want to advertise this behaviour until we've decided on how it should actually work. See #395, #396

milesfrain:
I suspect there's a lot of potential benefit for developers, even with the undefined behavior of `_`. Without it, there's no way to get these results.

What if an "experimental" note was added? (see recent PR update)

hdgarrood:
Thanks, but I don't want to add this note, I just want to fix the actual issue. The problem with adding the note is firstly that it's not experimental, it's just a bug arising from under-specification, and secondly, I think it will make it harder to change the way this works in the future, if people have gotten used to the current behaviour.

PR #418, opened 2020-09-06
Link to README in case of github rate limiting

### LINKED ISSUE #417: Link to README in case it can't be show due to rate limiting
Currently if there is the README can't be fetched due to rate limiting, pursuit is showing:

image

https://github.com/purescript/pursuit/blob/d995a0411c65aa4506074f10acc95188f550502b/src/TemplateHelpers.hs#L125-L129

## Possible improvement

We could add a hint ("Click here to open README instead") and additionally link to 
`https://<REPO>/blob/master/README.md`

andys8:
Tiny improvement to show hint and link to github in case the README couldn't be loaded due to rate limits.

Did not link to the README file directly, because wrong casing or ending would result in 404 (e.g. `README.md`, `Readme.md`, `readme.markdown`).

Closes #417

### DIFF
--- a/src/TemplateHelpers.hs
+++ b/src/TemplateHelpers.hs
@@ -4,6 +4,7 @@ import Import hiding (span, link)
 import qualified Data.List.Split as List
 import qualified Data.Map as Map
 import Data.Traversable (for)
+import Data.Bifunctor (first)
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks good!

PR #428, opened 2020-12-23
Update Pursuit for PureScript 0.14

thomashoneyman:
This PR updates Pursuit to compile with PureScript 0.14. There are still some checks that need to happen:

- [x] Verify that 0.13.x compilers can successfully upload to this version of Pursuit
- [x] Verify that 0.14.x compilers can successfully upload to this version of Pursuit
- [x] Verify that defaulting to `kindType` in `extractChildDeclarationType` doesn't break anything
- [x] Verify the [current Pursuit database](https://github.com/purescript/pursuit-backups) can be understood by this version of Pursuit

### DIFF
--- a/src/SearchIndex.hs
+++ b/src/SearchIndex.hs
@@ -277,6 +277,8 @@ extractChildDeclarationType declTitle declInfo cdeclInfo =
             , P.constraintArgs = map (P.TypeVar () . fst) args
             , P.constraintData = Nothing
             , P.constraintAnn = ()
+              -- TODO: Verify that defaulting to `kindType` works
+            , P.constraintKindArgs = map (fromMaybe (P.kindType $> ()) . snd) args
             }
         in
           Just (addConstraint constraint ty)
@@ -470,7 +472,8 @@ compareQual (P.Qualified _ a1) (P.Qualified _ a2) = a1 == a2
 
 runParser :: CST.Parser a -> Text -> Maybe a
 runParser p =
-  hush
+  fmap snd
+    . hush
     . CST.runTokenParser (p <* CSTM.token CST.TokEof)
     . CST.lexTopLevel
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,16 +1,28 @@
 resolver: lts-13.26
+# Resolves hjsmin requiring language-javascript ==0.6, while the purescript
+# compiler requires language-javascript ==0.7.
+allow-newer: true 
 packages:
 - '.'
 extra-deps:
 - bytestring-trie-0.2.5.0
+- cborg-0.2.2.0
 - connection-0.3.1
 - happy-1.19.9
-- language-javascript-0.6.0.13
+- language-javascript-0.7.0.0
 - network-3.0.1.1
-- purescript-0.13.3
+- ../purescript
+- ../purescript/lib/purescript-ast
+- ../purescript/lib/purescript-cst
+- serialise-0.2.2.0
 - socks-0.6.0
+- these-1.0.1
+- semialign-1
 flags:
   pursuit:
     dev: true
   aeson-pretty:
     lib-only: true
+  these:
+    assoc: false
+    quickcheck: false
--- a/stack.yaml
+++ b/stack.yaml
@@ -11,13 +11,17 @@ extra-deps:
 - happy-1.19.9
 - language-javascript-0.7.0.0
 - network-3.0.1.1
-- ../purescript
-- ../purescript/lib/purescript-ast
-- ../purescript/lib/purescript-cst
 - serialise-0.2.2.0
 - socks-0.6.0
 - these-1.0.1
 - semialign-1
+# purescript 0.14.0-rc4
+- github: purescript/purescript
+  commit: 8663d00a27c2c36e229f3d2f98358e660c297335
+  subdirs:
+    - .
+    - lib/purescript-ast
+    - lib/purescript-cst
 flags:
   pursuit:
     dev: true
--- a/pursuit.cabal
+++ b/pursuit.cabal
@@ -103,6 +103,7 @@ library
                  , vector
                  , time
                  , purescript
+                 , purescript-cst
                  , bower-json
                  , blaze-builder
                  , blaze-markup
--- a/src/Foundation.hs
+++ b/src/Foundation.hs
@@ -5,6 +5,7 @@
 module Foundation where
 
 import Import.NoFoundation
+import Language.PureScript.CoreFn.FromJSON (parseVersion')
 import Text.Read (readsPrec)
 import qualified Data.Text as T
 import qualified Data.Text.Lazy as LT
@@ -44,7 +45,7 @@ newtype PathVersion =
 
 instance PathPiece PathVersion where
   toPathPiece = toPathPiece . showVersion . runPathVersion
-  fromPathPiece = fmap PathVersion . D.parseVersion' . T.unpack
+  fromPathPiece = fmap PathVersion . parseVersion' . T.unpack
 
 -- | A base64 encoded string.
 newtype VerificationKey =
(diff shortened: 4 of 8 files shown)


stack.yaml
```diff
@@ -1,16 +1,28 @@
...
+- cborg-0.2.2.0
 - connection-0.3.1
 - happy-1.19.9
-- language-javascript-0.6.0.13
+- language-javascript-0.7.0.0
 - network-3.0.1.1
-- purescript-0.13.3
+- ../purescript
+- ../purescript/lib/purescript-ast
+- ../purescript/lib/purescript-cst
```
thomashoneyman, line 16: I tried to point this to GitHub following the [yaml configuration section of the Stack docs](https://docs.haskellstack.org/en/stable/yaml_configuration), but ultimately the build fails with `purescript` considered a missing dependency. I'm not sure how to point these dependencies somewhere other than local paths for testing's sake.

```yml
# purescript 0.14.0-rc4
- github: purescript/purescript
  commit: 8663d00a27c2c36e229f3d2f98358e660c297335
  subdirs:
    - lib/purescript-ast
    - lib/purescript-cst
```

src/SearchIndex.hs
```diff
@@ -277,6 +277,8 @@ extractChildDeclarationType declTitle declInfo cdeclInfo =
             , P.constraintArgs = map (P.TypeVar () . fst) args
             , P.constraintData = Nothing
             , P.constraintAnn = ()
+              -- TODO: Verify that defaulting to `kindType` works
+            , P.constraintKindArgs = map (fromMaybe (P.kindType $> ()) . snd) args
```
thomashoneyman, line 281: From @hdgarrood wrt testing this out:

> The places I'd be worried about displaying incorrect information are classes where kinds aren't explicitly declared but will be inferred as something other than Type (e.g. Functor), or where kinds are specified in a kind signature but not in a class declaration

For example:

```purs
class X :: (Type -> Type) -> Constraint
class X a
```

thomashoneyman:
@hdgarrood I've been able to verify that things look OK with this version of Pursuit. Below are some steps to reproduce the tests I ran through.

#### Verify existing database

First I verified that the updated Pursuit can load the entire set of Pursuit backups. I cloned the whole database into `data/verified` and made sure I could click around the modules:

Pursuit loads database

The packages I inspected manually look just fine, but I can check particular ones if you'd like.

#### Verify publishing with 0.13.8 and 0.14.0 compilers

First, I applied this patch to Pulp to help me test publishing packages to the local instance of Pursuit implemented in this pull request:

```diff
diff --git a/src/Pulp/Publish.purs b/src/Pulp/Publish.purs
index 77cd9dd..af5e2d2 100644
--- a/src/Pulp/Publish.purs
+++ b/src/Pulp/Publish.purs
@@ -309,7 +309,7 @@ readTokenFile = do
 
 pursuitUrl :: String -> Version -> String
 pursuitUrl name vers =
-  "https://pursuit.purescript.org/packages/" <> name <> "/" <> Version.showVersion vers
+  "http://localhost:3000/packages/" <> name <> "/" <> Version.showVersion vers
 
 uploadPursuitDocs :: Outputter -> String -> Buffer -> Aff Unit
 uploadPursuitDocs out authToken gzippedJson = do
@@ -331,8 +331,9 @@ uploadPursuitDocs out authToken gzippedJson = do
 
   reqOptions = fold
     [ HTTP.method := "POST"
-    , HTTP.protocol := "https:"
-    , HTTP.hostname := "pursuit.purescript.org"
+    , HTTP.protocol := "http:"
+    , HTTP.hostname := "localhost"
+    , HTTP.port := 3000
     , HTTP.path := "/packages"
     , HTTP.headers := headers
     ]
```

and ran `npm run build` to get a new `pulp.js` I could execute. Then I deleted the `prelude` package:

Verify prelude is gone

Then I verified I can publish the package with 0.13.8:

```sh
λ purs --version
0.13.8

λ node ../pulp.js publish --no-push
Checking your package is registered in purescript/registry... ok
Publishing purescript-prelude at v4.1.2. Is this ok? [y/n] y
* Uploading documentation to Pursuit...
* Done.
* You can view your package's documentation at: http://localhost:3000/packages/purescript-prelude/4.1.2
```

and made sure that this package ends up in the local `/data/verified`. It does get recorded with 'compilerVersion: 0.14.0' despite using the 0.13.8 compiler for the upload, but I suppose that's Pursuit annotating based on its library dependency on PureScript.

```sh
λ tail -c 27 data/verified/purescript-prelude/4.1.2.json | head -c 26
"compilerVersion":"0.14.0"
```

Here's the output:
Published with 0 13 8

I can also publish with 0.14.0-rc3:

```sh
λ purs --version                   
0.14.0-rc3

λ node ../pulp.js publish --no-push
Checking your package is registered in purescript/registry... ok
Publishing purescript-prelude at v4.1.4. Is this ok? [y/n] y
* Uploading documentation to Pursuit...
* Done.
* You can view your package's documentation at: http://localhost:3000/packages/purescript-prelude/4.1.4
```

and again verify the package ends up in `data/verified`.

```sh
λ tail -c 27 data/verified/purescript-prelude/4.1.4.json | head -c 26
"compilerVersion":"0.14.0"
```

Once again, the output:
Published with 0 14 0-rc4

#### Test possible edge cases

I tested your code snippet here by adding it to a new module in `prelude`:

```purs
module Prelude2 where

class X :: (Type -> Type) -> Constraint
class X a
```

and it seems to render just fine:
Test class renders correctly

though we're not rendering the kind annotations anywhere -- I assume this would be a nice thing to add, though.

stack.yaml
thomashoneyman: CI won't build until this is resolved, but I've been able to test the 4 things we wanted to test locally and everything checks out.

src/SearchIndex.hs
thomashoneyman: I've had a chance to test this out and it seems to work just fine -- though we aren't displaying the kind signatures in the documentation.

hdgarrood:
Yes, we should really update the HTML documentation generation inside the compiler soon so that code taking advantage of PolyKinds is documented accurately.

That all sounds great apart from one thing:

> and made sure that this package ends up in the local /data/verified. It does get recorded with 'compilerVersion: 0.14.0' despite using the 0.13.8 compiler for the upload, but I suppose that's Pursuit annotating based on its library dependency on PureScript.

Pursuit shouldn’t be doing this, the compiler version in the uploaded JSON should contain the compiler version used to create that JSON originally. Are you absolutely sure you used 0.13.8 to produce that JSON? If Pursuit is overwriting that field then that’s a bug we should fix.

stack.yaml
hdgarrood: What happens if you add a “.” entry in the subdirs list?

thomashoneyman:
> Pursuit shouldn’t be doing this, the compiler version in the uploaded JSON should contain the compiler version used to create that JSON originally. Are you absolutely sure you used 0.13.8 to produce that JSON? If Pursuit is overwriting that field then that’s a bug we should fix.

I’m sure that the compiler in my PATH is 0.13.8 in the first section (which fails to build prelude on master), and 0.14.0-rc3 in the second section. I don’t have the compilers in the same location, so the actual commands I’m running in the second step first have:

    PATH=path/to/purs:$PATH <command>

hdgarrood:
Ah, I see what's going on. Currently the `ToJSON` instance for an uploaded package overwrites the current compiler version when encoding to JSON, see https://github.com/purescript/purescript/blob/d056c120821cc11f654cf15071cc0b43e98463f8/src/Language/PureScript/Docs/Types.hs#L758. This will need to be fixed inside the compiler so it's not worth holding this change up for it, I think.

hdgarrood (approved):

PR #430, opened 2021-01-21
Default constraintKindArgs to an empty list

thomashoneyman:
In `constraintKindArgs` we previously defaulted any missing kind annotation to `kindType`, which ended up filling missing kind annotations with an application `@Type`. This is now set to an empty list.

I've also bumped the version in the .cabal file so this can be released. I can do that in a separate commit, though, if desired.

### DIFF
--- a/src/SearchIndex.hs
+++ b/src/SearchIndex.hs
@@ -277,7 +277,7 @@ extractChildDeclarationType declTitle declInfo cdeclInfo =
             , P.constraintArgs = map (P.TypeVar () . fst) args
             , P.constraintData = Nothing
             , P.constraintAnn = ()
-            , P.constraintKindArgs = fold $ traverse snd args
+            , P.constraintKindArgs = []
             }
         in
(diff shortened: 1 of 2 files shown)


src/SearchIndex.hs
```diff
@@ -277,8 +277,7 @@ extractChildDeclarationType declTitle declInfo cdeclInfo =
             , P.constraintArgs = map (P.TypeVar () . fst) args
             , P.constraintData = Nothing
             , P.constraintAnn = ()
-              -- TODO: Verify that defaulting to `kindType` works
-            , P.constraintKindArgs = map (fromMaybe (P.kindType $> ()) . snd) args
+            , P.constraintKindArgs = fold $ traverse snd args
```
natefaubion: This should just be `[]`.

thomashoneyman: Done in 5f03e4f.

natefaubion (approved):

PR #437, opened 2021-07-07
Add documentation on kind signatures, which ones are displayed, and why

### LINKED ISSUE #434: Add an explanation on kind signatures appearing in documentation
Follow up from this comment: https://github.com/purescript/purescript/pull/4119#issuecomment-869022915

Below is what I originally wrote for the changelog entry:

---

Previously, data/newtype/type/class declarations that have explicit kind
signatures would not display those kind signatures in their documentation.
For example, the two below types...

```purescript
data PolyProxy :: forall k. k -> Type
data PolyProxy a = PolyProxy

data TypeProxy :: Type -> Type
data TypeProxy a = TypeProxy
```

... would only show the following information in their docs. One cannot
be distinguished from another due to the missing kind signatures:

```
data PolyProxy a = PolyProxy

data TypeProxy a = TypeProxy
```

Now, these types' kind signatures are displayed above their declarations
in their docs, similar to what one would see in the source code. The only
exception to this are kind signatures whose kinds are "uninteresting"
(covered below).

Moreover, declarations without explicit kind signatures will still
display the kind signatures inferred by the compiler. For example,
`Foo`'s documentation will display the below (commented-out) inferred
kind signature:

```purescript
                                                {-
data Foo :: (Type -> Type) -> Type -> Type      -}
data Foo f a = Foo (f Int) a
```

However, a kind signature for both explicit and inferred kinds will not
be displayed if that kind is considered "uninteresting." An "uninteresting"
kind is one that follows this form:
- `Type`
- `Constraint`
- `Type -> K` where `K` refers to an "uninteresting" kind signature

Thus, both the `Bar` and `Baz` declarations below will not have their
explicit and inferred kind signatures displayed in their docs.
```purescript
data Bar :: Type -> Type -> Type
data Bar a b = Bar a b
                                                  {-
class Baz :: Type -> Type -> Constraint           -}
class Baz a b where
  doBaz :: a -> b -> String
```

JordanMartinez:
Fixes #434 

I'll have to update the changelog with an entry for this after #436 gets merged.

### DIFF
--- a/static/help-docs/users.md
+++ b/static/help-docs/users.md
@@ -79,3 +79,125 @@ called "Kleisli" in some contexts. This may also change in the future; see
 [`sequence`]: https://pursuit.purescript.org/packages/purescript-foldable-traversable/docs/Data.Traversable#t:Traversable
 [Modules page in the documentation repository]: https://github.com/purescript/documentation/blob/master/language/Modules.md
 [`Star`]: https://pursuit.purescript.org/packages/purescript-profunctor/docs/Data.Profunctor.Star#t:Star
+
+## Kind Signatures
+
+### Explicit and Inferred
+
+Data, newtype, type synonym, and type class declarations all have kind
+signatures. These signatures are either explicit (i.e. developer-defined)
+or implicit (i.e. compiler-inferred). For example
+
+```purescript
+-- Explicit kind signature
+
+data ExplicitFoo :: forall k. k -> Type
+data ExplicitFoo a = ExplicitFoo
+
+-- Implicit kind signature
+                                                                  {-
+data ImplicitFoo :: forall k. k -> (Type -> Type) -> Type         -}
+data ImplicitFoo a f = ImplicitFoo (f Int)
+```
+
+### Merging Documentation Comments
+
+Since both the kind signature declaration and the data/newtype/type/class
+declaration can have documentation comments, both will be merged together
+with a newline separating them, displaying the kind signature declaration's
+documentation comments first and then the other declaration's documentation
+comments second. For example, the below code...
+```purescript
+-- | Kind signature declaration documentation comment
+data ExplicitFoo :: forall k. k -> Type
+-- | Data declaration documentation comment
+data ExplicitFoo a = ExplicitFoo
+```
+... will be rendered as
+```
+Kind signature declaration documentation comment
+Data declaration documentation comment
+```
+
+### Interesting Kinds are displayed; Uninteresting Kinds are not
+
+The following design choice should make it easier for new learners
+to learn the language by limiting their exposure to these more
+advanced language features.
+
+In some cases, kind signatures can be very helpful. In other cases, they are
+a needless distraction. By default, all kind signatures, whether explicitly-
+declared by the developer or inferred by the compiler, will only be displayed
+if the kind signatures are considered "interesting." Put differently,
+"uninteresting" kind signatures will not be displayed.
+
+An "uninteresting" kind signature is one that follows this form:
+- `Type`
+- `Constraint`
+- `Type -> K` where `K` refers to an "uninteresting" kind signature
+
+Here's another way to think about it. Kind signatures are considered
+"uninteresting" if all of their type parameters' kinds are kind `Type`.
+
+### Examples of "uninteresting" kind signatures
+
+Consider the following examples of "uninteresting" kind signatures. Each
+kind signature is considered "uninteresting" because
+1. it does not have type parameters
+2. if it has type parameters, each type parameters' kind is kind `Type`
+
+```purescript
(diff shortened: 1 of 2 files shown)


static/help-docs/users.md
```diff
@@ -79,3 +79,125 @@ called "Kleisli" in some contexts. This may also change in the future; see
...
+```purescript
+-- Explicit kind signature
+
+data ExplicitFoo :: forall k. k -> Type
+data ExplicitFoo a = ExplicitFoo
+
+-- Implicit kind signature
+                                                                  {-
+data ImplicitFoo :: forall k. k -> (Type -> Type) -> Type         -}
+data ImplicitFoo a f = ImplicitFoo (f Int)
```
thomashoneyman, line 100: Perhaps we could do something to make it a little clearer the signature is being inferred by something like:

```purs
{- The compiler will infer this kind signature for the below type:
data ImplicitFoo :: forall k. k -> (Type -> Type) -> Type
-}
data ImplicitFoo a f = ImplicitFoo (f Int)

File: static/help-docs/users.md [hunk 2]
```diff
@@ -79,3 +79,125 @@ called "Kleisli" in some contexts. This may also change in the future; see
...
+-- | Data declaration documentation comment
+data ExplicitFoo a = ExplicitFoo
+```
+... will be rendered as
+```
+Kind signature declaration documentation comment
+Data declaration documentation comment
+```
+
+### Interesting Kinds are displayed; Uninteresting Kinds are not
```
thomashoneyman, line 122: I don't think "Kind" has to be capitalized

```suggestion
### Interesting kinds are displayed; uninteresting kinds are not
```

static/help-docs/users.md
```diff
@@ -79,3 +79,125 @@ called "Kleisli" in some contexts. This may also change in the future; see
...
+if the kind signatures are considered "interesting." Put differently,
+"uninteresting" kind signatures will not be displayed.
+
+An "uninteresting" kind signature is one that follows this form:
+- `Type`
+- `Constraint`
+- `Type -> K` where `K` refers to an "uninteresting" kind signature
+
+Here's another way to think about it. Kind signatures are considered
+"uninteresting" if all of their type parameters' kinds are kind `Type`.
```
thomashoneyman, line 140:
```suggestion
Here's another way to think about it: kind signatures are considered
"uninteresting" if all of their type parameters' kinds are of kind `Type`.
```

static/help-docs/users.md
```diff
@@ -79,3 +79,125 @@ called "Kleisli" in some contexts. This may also change in the future; see
...
+
+-- every type parameter makes this kind signature "interesting"
+class TypeLevelProgrammingFunction :: Symbol -> Row Type -> Row Type -> Constraint
+class TypeLevelProgrammingFunction sym row1 row2 | sym row1 -> row2
+```
+
+In the `PolyProxy` example, the polykinded `a` type parameter is interesting.
+
+In the `FunctorLike` and `BiFunctorLike` examples, the higher-kinded `f`
+type paramter is interesting.
```
thomashoneyman, line 201:
```suggestion
In the `FunctorLike` and `BiFunctorLike` examples, the higher-kinded `f`
type parameter is interesting.
```

thomashoneyman (commented):
I've included a few comments. In general this feels a little long; I wonder if we could trim some of the details or move them into the `purs docs` documentation instead.

I'm also not totally sold on the 'interesting' vs. 'uninteresting' terminology, though I don't have a better pairing in mind at the moment.

JordanMartinez:
I've addressed all feedback. I tried making it shorter by removing a few things. How is it now?

kl0tl (commented):
The example sections feel redundant with the “Interesting kinds are displayed; Uninteresting kinds are not” section. Perhaps we could provide examples when explaining what kinds are uninteresting and then show some interesting kinds as counterexamples?

JordanMartinez:
> Perhaps we could provide examples when explaining what kinds are uninteresting and then show some interesting kinds as counterexamples?

Isn't that what I already do in this PR in the subsections?

static/help-docs/users.md
JordanMartinez: @kl0tl Here's what I was doing above originally. Thomas is the one who suggested the edit that you commented on below.

Which is preferred?

kl0tl:
Yes, but I don’t think it is necessary to rehash what uninteresting kinds are in the “Examples of "uninteresting" kind signatures“ section.

JordanMartinez:
How are the latest changes?

thomashoneyman (approved):

JordanMartinez:
@thomashoneyman Thanks for the approval! This has been merged. We can redeploy Pursuit and finally announce v0.14.3

PR #464, opened 2022-05-06
Update `purescript` dep to 0.15.0

JordanMartinez:
I tried regenerating the `LICENSE` file but don't have https://hackage.haskell.org/package/cabal-dependency-licenses installed. I'll need to figure that out before we merge and redeploy this.


JordanMartinez:
Actually, it looks like that license library is no longer maintained, but [it does recommend an alternative](https://github.com/jaspervdj/cabal-dependency-licenses#cabal-dependency-licenses=): 

[cabal-plan](https://hackage.haskell.org/package/cabal-plan#description)

thomashoneyman (approved):

JordanMartinez:
@hdgarrood Just a few questions...
- are my license generation changes valid? I couldn't get `cabal-dependency-licenses` to build, so I had to use `cabal-plan`.
- do we still need the LGPL section?

hdgarrood:
The changes to LICENSE seem reasonable to me. If we aren’t depending on anything that’s LGPL any more then I think we can get rid of that section.

JordanMartinez:
Cool. I'll get rid of it.

JordanMartinez:
Pursuit has now been deployed.

PR #468, opened 2022-08-22
Update to v0.15.4

sigma-andex:
Update pursuit to 0.15.4. 

Apparently `barrier`, the library to render badges, is hopelessly outdated and doesn't work with GHC 9.2.3. I haven't found any replacement for rendering badges. Not sure if anyone wants to fork, update and maintain this library? Otherwise I would propose to drop badge support.


sigma-andex:
@PureFunctor could you check if my updates in `src/SearchIndex.hs` are correct?

src/SearchIndex.hs
```diff
@@ -297,7 +297,7 @@ extractChildDeclarationType declTitle declInfo cdeclInfo =
 
   where
     parentName :: P.Qualified (P.ProperName a)
-    parentName = P.Qualified Nothing (P.ProperName declTitle)
+    parentName = P.Qualified (P.ByNullSourcePos) (P.ProperName declTitle)
```
PureFunctor, line 300: Aside from the redundant parentheses, this LGTM

```suggestion
    parentName = P.Qualified P.ByNullSourcePos (P.ProperName declTitle)
```

config/routes
```diff
@@ -10,7 +10,6 @@
 /packages/#PathPackageName/#PathVersion/docs           PackageVersionDocsR         GET
 /packages/#PathPackageName/#PathVersion/docs/#Text     PackageVersionModuleDocsR   GET
 !/packages/#PathPackageName/docs/#Text                 PackageModuleDocsR          GET
-!/packages/#PathPackageName/badge                      PackageBadgeR               GET
```
PureFunctor, line 13: I'm not actually sure we can drop badge support altogether. If I recall correctly, most of the packages within core make use of this for their READMEs. If we need to fork `badger` to work with 9.2.3 then I could probably pick up that task.

PureFunctor (changes requested):

config/routes
PureFunctor: Looks like it just needs its upper bound updated: https://github.com/PureFunctor/barrier/commit/79e8fc36fb7d36591f9f65c0e0eb892d8062cdc5

sigma-andex: cool, yeah that sounds doable.

JordanMartinez: Looks like Phil did the same thing a few years ago: https://github.com/philopon/barrier/pull/4

sigma-andex: ok I guess we can try open a PR and see if we get a response

sigma-andex: Ok I undid the badge removal and added @PureFunctor 's fork/commit as a local dependency. This builds for me using cabal. I'm not sure if it works on stack, as I cannot get stack to run. Would be great if one of you could test it.
Or do we actually need/use stack?

PureFunctor: Yeah, we need to specify a git dependency for it under `extra-deps`. I don't have a `stack` built at hand at the moment, since it's pretty wonky at M1. I will transfer over my fork to the `purescript` organization in the meantime as well.

sigma-andex: I added it to extra-deps, but I can't get it to build.

PureFunctor: Looking into it...

PureFunctor: Ok, I think I need to relax the bounds for `bytestring` in `barrier`: PureFunctor/barrier@db5e27c

sigma-andex: Thanks to @i-am-the-slime I got the stack build working as well.

sigma-andex: @PureFunctor if you move it to `purescript-contrib` I can update the dependencies. Apart fromt his I think it is good to go now

src/Handler/Caching.hs
```diff
@@ -102,7 +102,6 @@ clearCache pkgName version = do
       eachRouteDir
         [ PackageVersionR pkgName' (PathVersion version)
         , PackageAvailableVersionsR pkgName'
-        , PackageBadgeR pkgName'
```
JordanMartinez, line 105: Should this be readded?

stack.yaml
```diff
@@ -23,3 +26,6 @@ flags:
     lib-only: true
   these:
     assoc: false
+system-ghc: true
+compiler: ghc-9.2.3
+compiler-check: match-exact
```
JordanMartinez, line 31: I'm not familiar with `stack.yaml` in general. Why are these needed?

src/Handler/Caching.hs
sigma-andex: uh yes

stack.yaml
sigma-andex: oh yes that accidenily slipped into the commit

src/Handler/Caching.hs
sigma-andex: Fixed

stack.yaml
sigma-andex: Fixed

JordanMartinez (approved):

PureFunctor:
Woops, CI breaks because I force pushed on my `barrier` fork 😅. Updating the commit hash to PureFunctor/barrier@db5e27c should make this work.

sigma-andex:
Ok updated it to this commit

sigma-andex:
@JordanMartinez can you re-run the pipeline?

JordanMartinez:
> @JordanMartinez can you re-run the pipeline?

Restarted. Sorry about the delay.

sigma-andex:
Next try :D

sigma-andex:
Ok I think now it should be good :D

sigma-andex:
Once this is good, it would be great to get #469 approved & merged as well and then do a deployment

JordanMartinez:
@garyb Assuming this builds, can I get another approval here? I'm not sure if @PureFunctor's approval (if it was given) would count due to not being a core team member. If I'm wrong, feel free to correct me.

sigma-andex:
> @garyb Assuming this builds, can I get another approval here? I'm not sure if @PureFunctor's approval (if it was given) would count due to not being a core team member. If I'm wrong, feel free to correct me.

How about promoting @PureFunctor to the core team so you can be sure about it? :)

JordanMartinez (approved):

JordanMartinez:
Thanks all!

Issue #227, opened 2016-07-15
purescript-halogen is missing


gabysbrain:
The links to functions from the purescript-halogen wiki are dead and there's no mention of the halogen on the pursuit package list

paf31:
I would mention this on the Halogen repo instead. Someone just needs to reupload it.

Issue #215, opened 2016-05-08
purescript-halogen has been missing for a few days


stefanholzmueller:
Title basically says it all. For example, this link is now broken: https://pursuit.purescript.org/packages/purescript-halogen/0.6.1/docs/Halogen.HTML.Properties#v:class_

paf31:
This probably means the reupload script was unable to build the upload package, which probably means something isn't building correctly. If the library can be built with 0.8.5, then anyone should be able to reupload the JSON data.

hdgarrood:
Yeah, I think it was because halogen is too large without gzipping and my script doesn't do any gzipping.

@stefanholzmueller, I've recently released a version of pulp which should make it easy for anyone to upload docs. Can you please try cloning the halogen repository, checking out the tag you want (eg v0.6.1) and then running `pulp publish --no-push`? You'll need pulp 8.2.0 at least.

stefanholzmueller:
@hdgarrood The last time I upgraded pulp, it broke my whole setup. I do not want to risk that now, but I'll try it in a few days. Thanks for the advice though!

hdgarrood:
Never mind, I just uploaded it.

Issue #189, opened 2016-02-12
Halogen docs have disappeared


nuttycom:
This may not be quite the right place to ask about this, but it looks like the purescript-halogen docs, which I was using just a couple of days ago, have completely disappeared. Here's the missing link: https://pursuit.purescript.org/packages/purescript-halogen/0.5.14

hdgarrood:
It's as good a place as any.

Yes, the problem is that after the re-exports changes, halogen (and batteries) are just too big and nginx is rejecting them with Request Entity Too Large. That's http status code 413, IIRC.

The fix is one or both of #188 and configuring nginx to increase the maximum permitted request body size. We just haven't gotten around to it yet.

hdgarrood:
Sorry, that's supposed to be #186

hdgarrood:
Fixed: https://pursuit.purescript.org/packages/purescript-halogen/0.5.18

PR #471, opened 2022-09-20
Update nginx max body size to 10m

thomashoneyman:
As seen [in this failed workflow on the registry](https://github.com/purescript/registry/actions/runs/3088187056/jobs/4994371593), some package resolutions exceed the 1m default body size imposed by nginx, and so this PR raises that limit to be 10m instead.

Relates to #189, #215, #227, and so on – this has been an issue for some time.

### DIFF
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -25,6 +25,11 @@ server {
 
     listen 443 ssl http2;
     listen [::]:443 ssl http2;
+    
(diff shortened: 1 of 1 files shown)


JordanMartinez (approved):

thomashoneyman:
Oops — published without re-running this. I can remove the release since this isn’t used as a library.

PR #490, opened 2026-07-06
Harden package pages against scraper-driven heap exhaustion

thomashoneyman:
v0.9.10 (#486) and v0.9.11 (#487) handled the kernel OOM kills, but the server is still going down (albeit much less frequently): since the v0.9.11 deploy the process has died a few times with RTS heap-overflow aborts (systemd `status=251`) and frozen in GC thrash, tripping the `/search` uptime check each time.

The trigger is bursts of scrapers (~5000 distinct IPs in the last day, with spoofed Chrome/Edge user agents, crawling the full site). Crawl bursts produce the crashes. So this change institutes yet another set of checks.

Two changes, one per commit:

1. **Give large package decodes an aggregate memory budget.** Decodes of files of 1MB or more share a 16MB in-flight byte budget. A decode is admitted once the total size of large files being decoded fits within the budget, or immediately when it is the only large decode running (so files bigger than the budget, e.g. react-icons at 25MB, are still served, by themselves). The search index regeneration is exempted.

2. **Challenge browser-like scraper traffic to package pages with Anubis.** Deploys now install [Anubis](https://github.com/TecharoHQ/anubis), and nginx routes cache misses under `/packages/` through it. The basic idea is that clients with browser-like user agents have to solve a JavaScript proof of work challenge once and then carry a cookie for a week; non-browser clients like curl, package uploads, etc. are not checked. Cache hits, search, and post requests to packages do not go through anubis.


thomashoneyman:
Added one more commit after load-testing this branch against the full pursuit-backups dataset. Found a starvation problem where a file bigger than the whole budget (react-icons is 24MB) is only admitted when nothing else is in flight, and STM retries impose no arrival order — so under a steady stream of smaller large-file decodes, its waiters are overtaken indefinitely.

